PR #27, opened 2019-10-02
Fixed logging an error when retrieving kms secret succeeds

alijared:


Kav91:
Actually @alijared linter is failing, could you plz gofmt the the secrets file.
internal/config/secrets.go:266: File is not `gofmt`-ed with `-s` (gofmt)
 
https://buildkite.com/kav91/nri-flex/builds/178#2c3a85bb-7520-49c7-a8de-941483951630/246-253

PR #29, opened 2019-10-09
Func RunKeepKeys not deleting keys

heaje:
During some testing today I found that the "keep_keys" option didn't appear to be working properly.  After going through various test scenarios (and code), I found that the delete() statement in "RunKeepKeys()" appears to be using the wrong variable.   After switching to *key, the statement appears to do what I expected.


Kav91:
@heaje apologies for the delayed reply, thank you for catching and fixing this bug!

PR #31, opened 2019-10-23
Updated RunSampleRenamer - change the conflicting variable name regex…

haihongren:
- The RunSampleRenamer is not working properly due to conflicting variable name "regex".
- Changing the local "regex" variable name to "regexLocal" to avoid overriding the same variable defined on line 26.


Kav91:
👍

PR #56, opened 2020-01-16
add go mod modules support

ardias:
Add go mod modules support. Makes it a bit easier to prepare a local environment for testing.


Kav91:
@ardias @carlosroman this breaks make `package-all`.

ardias:
Sorry @Kav91 . I tested it with make and the CI build. I'll fix it.

PR #61, opened 2020-01-22
add cmd arg to copy nrjmx files during install

icpenguins:
Simple change to allow a command-line argument to be added when running the installation to copy the nrjmx files.


Kav91:
LGTM thanks for this!

PR #67, opened 2020-01-24
Added containerized test run and Travis integration

mariomac:
Allows executing all the tests (integration included) from a docker compose file.

It also enables travis

### DIFF
--- a/integration-test/docker-compose.yml
+++ b/integration-test/docker-compose.yml
@@ -1,0 +1,19 @@
+version: "3.1"
+services:
+  nri-flex-src:
+    build:
+      context: ..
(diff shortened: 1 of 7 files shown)


integration-test/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: "3.1"
```
cristianciutea: is this file in root still required? https://github.com/newrelic/nri-flex/blob/master/docker-compose.yml

mariomac: As long as BuildKite is activated, yes. We will remove it when we remove the BuildKite tests.

PR #71, opened 2020-02-03
Improve error handing

cristianciutea:
This PR improves the error handling


internal/config/git.go
```diff
@@ -110,16 +100,19 @@ func GitPull(dir string) error {
 	// instance\iate a new repository targeting the given path (the .git folder)
```
ardias: this should probably read '**instantiate** a new ...'

internal/config/git.go
```diff
@@ -91,16 +78,19 @@ func GitClone(dir string, u *url.URL) error {
 	})
 
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to clone repo, error: %v", err)
```
ardias: can you add the **config:**  prefix to the error messages?

internal/config/git.go
```diff
@@ -91,16 +78,19 @@ func GitClone(dir string, u *url.URL) error {
...
 	w, err := r.Worktree()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get working directory while cloning the repo, error: %v", err)
 	}
 
-	GitCheckout(w)
+	err = GitCheckout(w)
+	if err != nil {
+		return fmt.Errorf("failed to clone repo, error: %v", err)
```
ardias: same as above

internal/config/secrets.go
```diff
@@ -35,13 +35,13 @@ func loadSecrets(config *load.Config) {
 		if secret.Kind == "" {
 			load.Logrus.WithFields(logrus.Fields{
 				"secret": name,
-			}).Error("config: secret is missing kind")
+			}).Error("config: secret missing: 'kind'")
```
ardias: this makes it look like a secret named **kind** missing. I would keep the other phrase or write something like:
`config:  secret 'kind' is required` or `config:  secret 'kind' is missing`  or like you have below :
`config: secret needs 'kind' parameter to be set`

internal/config/secrets.go
```diff
@@ -60,7 +60,7 @@ func loadSecrets(config *load.Config) {
 				load.Logrus.WithFields(logrus.Fields{
 					"secret": name,
 					"kind":   secret.Kind,
-				}).Error("config: secret missing region")
+				}).Error("config: secret missing: 'region'")
```
ardias: I would use the same format as below: `.. requires 'region' parameter to be set`

internal/config/secrets.go
```diff
@@ -323,7 +323,7 @@ func httpWrapper(secret load.Secret) ([]byte, error) {
 		if err != nil {
 			load.Logrus.WithFields(logrus.Fields{
 				"err": err,
-			}).Error("config: secret failed to read ca")
+			}).Error("config: secret failed to read tls ca")
```
ardias: why not use `WithError(err).Error(...)` ?

internal/outputs/http.go
```diff
@@ -53,20 +44,16 @@ func postRequest(url string, key string, data []byte) {
...
-	if resp != nil {
-		if resp.StatusCode > 299 || resp.StatusCode < 200 {
-			load.Logrus.WithFields(logrus.Fields{
-				"code": resp.StatusCode,
-			}).Error("http: post failed")
-		}
-	} else {
-		load.Logrus.Error("http: response nil")
+	if resp == nil {
+		return errors.New("http: response nil")
```
ardias: why do we use **errors.New** here instead of **fmt.Errorf**. Even if we don't anything to format, it's strange to use a different way to create the error

internal/outputs/infra.go
```diff
@@ -6,47 +6,49 @@
...
+	load.Entity, err = createEntity(load.Args.Local, load.Args.Entity)
 	if err != nil {
-		load.Logrus.WithFields(logrus.Fields{"err": err}).Fatal("flex: create integration")
-	}
-
-	if load.Args.Local {
-		load.Entity = load.Integration.LocalEntity()
-	} else {
-		InfraRemoteEntity()
+		return fmt.Errorf("flex: failed create remote entity: %v", err)
```
ardias: in this case it should be just "failed to create entity", because it can be a local or remote entity, right?

internal/outputs/insights.go
```diff
@@ -9,17 +9,16 @@ import (
...
 
-	"github.com/newrelic/nri-flex/internal/load"
-	"github.com/sirupsen/logrus"
-
 	"github.com/newrelic/infra-integrations-sdk/data/metric"
 	"github.com/newrelic/infra-integrations-sdk/integration"
+	"github.com/newrelic/nri-flex/internal/load"
 )
 
 // SendToInsights - Send processed events to insights
```
ardias: this comment does not belong here

cmd/nri-flex/nri-flex.go
```diff
@@ -9,24 +9,28 @@ import (
...
 		// default process
 		integration.SetDefaults()
-		integration.RunFlex("")
+		integration.RunFlex(integration.FlexModeDefault)
 	}
 
-	err := load.Integration.Publish()
-	if err != nil {
-		load.Logrus.WithFields(logrus.Fields{"err": err}).Fatal("flex: unable to publish")
+	if err := load.Integration.Publish(); err != nil {
```
ardias: I would leave the check for error like it was, instead of putting everything in one line. It's easier to read and it's also how it's done in most all other files

internal/config/git.go
cristianciutea: is added in the logging (in the place where is called this function

PR #73, opened 2020-02-04
adias/add linux command integration test

ardias:


internal/config/fetch_test.go
```diff
@@ -1,75 +1,73 @@
 package config
 
-import (
-	"testing"
-)
+// TODO: this test is flaky. Should be revised
```
davidbrota: please, make sure this test will be revised in some other task

internal/load/README.md
```diff
@@ -192,7 +193,7 @@ type API struct {
 	Events            map[string]string `yaml:"events"`         // set as events
 	EventsOnly        bool              `yaml:"events_only"`    // only generate events
 	Merge             string            `yaml:"merge"`          // merge into another eventType
-	Joinkey           string            `yaml:"joinkey"`        // merge into another eventType
+	JoinKey           string            `yaml:"join_key"`       // merge into another eventType
```
davidbrota: Doesn't this change break the existing config files using the old key? maybe we should add a comment in the changelog (just in case)

internal/load/README.md
```diff
@@ -289,6 +290,7 @@ type ArgumentList struct {
 	ContainerDiscoveryDir   string `default:"flexContainerDiscovery/" help:"Set directory of auto discovery config files"`
 	ContainerDiscovery      bool   `default:"false" help:"Enable container auto discovery"`
 	ContainerDiscoveryMulti bool   `default:"false" help:"Allow a container to be matched multiple times"`
+	ContainerDump           bool   `default:"false" help:"Dump all containers, useful for debugging"`
```
davidbrota: Is this field being used in this PR?

internal/load/README.md
ardias: This is just the documentation. It seems it was changed in the code, but not in this readme file

internal/load/README.md
ardias: This is documentation. It's auto generated so it seems it was missing from the readme

internal/config/fetch_test.go
ardias: Ok, i've created a ticket in our issue tracker

internal/load/README.md
davidbrota: oh, my bad. I didn't see both changes were made in a README 🤦‍♂

PR #77, opened 2020-02-08
feat: output http errors to sample

### LINKED ISSUE #75: Send an event to Insights when http request fails
Right now, an HTTP request failure is only visible in the form of debug statements that can only be seen when Flex is run on verbose mode.

https://github.com/newrelic/nri-flex/blob/master/internal/inputs/http.go#L151
This appears to be where http error handling logic resides. I'm wondering if this is a good place to add additional logic to push http errors to Insights for users who may want to alert off of failed http requests. This is a common use case for users who use Flex for http health checks.

Kav91:
When a connection failure occurs, no sample is generated.

This change will output those errors into a sample.
```json
{
    "error": "Get https://asda.asdfadf331.com/blah: dial tcp: lookup asda.asdfadf331.com: no such host",
    "event_type": "httpSampleABC",
    "integration_name": "com.newrelic.nri-flex",
    "integration_version": "Unknown-SNAPSHOT"
}
```

Fixes #75


mariomac:
LGTM, temporarily removing the approval as the teams requires discussing the approach as an stardar for other similar errors.

PR #78, opened 2020-02-11
fix: remove variableProcessor from loop, and make available for lookups

Kav91:
- runVariableProcessor should not run within loop
- runVariableProcessor should be functional when lookups are used


haihongren:
@Kav91 
Moving runVariableProcessor out of the loop appears to break the following use case: 
https://github.com/newrelic/nri-flex/blob/master/examples/flexConfigs/variable-store-example.yml 

Please validate it on your end as well.

PR #79, opened 2020-02-11
adias/return exit errorcodes

ardias:
Return error code when flex fails to start/run due to configuration file issues.
Also added an option to print log messages in json structure so that external tools can better parse them

### DIFF
--- a/internal/integration/integration.go
+++ b/internal/integration/integration.go
@@ -6,6 +6,7 @@
 package integration
 
 import (
+	"fmt"
 	"github.com/newrelic/nri-flex/internal/utils"
 	"io/ioutil"
 	"os"
@@ -34,7 +35,7 @@ const (
 )
 
 // RunFlex runs flex.
-func RunFlex(runMode FlexRunMode) {
+func RunFlex(runMode FlexRunMode) error {
 	setupLogger()
 
 	load.Logrus.WithFields(logrus.Fields{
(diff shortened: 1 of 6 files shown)


internal/integration/integration.go
```diff
@@ -48,31 +49,48 @@ func RunFlex(runMode FlexRunMode) {
 
 	switch runMode {
 	case FlexModeLambda:
-		addConfigsFromPath("/var/task/pkg/flexConfigs/", &configs)
+		errors := addConfigsFromPath("/var/task/pkg/flexConfigs/", &configs)
+		if len(errors) > 0 {
+			return fmt.Errorf("flex: failed to read some configuration files. please review them before continuing")
```
Kav91: I would use a comma instead of period, since we are not ending the sentence

internal/integration/integration.go
```diff
@@ -48,31 +49,48 @@ func RunFlex(runMode FlexRunMode) {
...
 
 		isSyncGitConfigured, err := config.SyncGitConfigs("/tmp/")
 		if err != nil {
-			logrus.WithError(err).Error("flex: failed to sync git configs")
+			logrus.WithError(err).Warn("flex: failed to sync git configs")
 		} else if isSyncGitConfigured {
-			addConfigsFromPath("/tmp/", &configs)
+			errors = addConfigsFromPath("/tmp/", &configs)
+			if len(errors) > 0 {
+				logrus.WithError(err).Warn("flex: failed to load git sync configuration files. ignoring and continuing")
```
Kav91: same with here

internal/integration/integration.go
Kav91: also should this not also be logrus.WithError()?

ardias: this is one of those cases where I was not sure if we should stop processing if we encounter an "unrecoverable" in one of the files we're trying to load.
If we continue we (the user) may not notice it and question themselves why one of his configurations is not reporting.
If you're feeling is that we should continue lmk and I'll change the behaviour, but I do think having a "healthy" config folder is better than having some random file failing

PR #80, opened 2020-02-11
chore: improve content header check

Kav91:
- use equals instead of contains for json contentType check
- add xml contenType check (without this when it falls through the switch and attempts to check the payload, the XML check is not always reliable when the `<xml` tag is not present.)

### DIFF
--- a/internal/inputs/http.go
+++ b/internal/inputs/http.go
@@ -92,7 +92,7 @@ func RunHTTP(dataStore *[]interface{}, doLoop *bool, yml *load.Config, api load.
 			case contentType == "text/xml" || contentType == "application/xml":
 				jsonBody, err := xj.Convert(resp.Body)
 				if err != nil {
-					load.Logrus.WithError(err).Error("http: URL %v failed to convert XML to Json resp.Body", *reqURL)
+					load.Logrus.WithError(err).Errorf("http: URL %v failed to convert XML to Json resp.Body", *reqURL)
 				} else {
 					if api.Pagination.OriginalURL == "" || (api.Pagination.OriginalURL != "" && resp.StatusCode >= 200 && resp.StatusCode <= 299) {
 						handleJSON(dataStore, jsonBody.Bytes(), &resp, doLoop, reqURL, nextLink)
--- a/internal/inputs/http.go
+++ b/internal/inputs/http.go
@@ -92,9 +92,7 @@ func RunHTTP(dataStore *[]interface{}, doLoop *bool, yml *load.Config, api load.
 			case contentType == "text/xml" || contentType == "application/xml":
 				jsonBody, err := xj.Convert(resp.Body)
 				if err != nil {
-					load.Logrus.WithFields(logrus.Fields{
-						"err": err,
-					}).Errorf("http: URL %v failed to convert XML to Json resp.Body", *reqURL)
+					load.Logrus.WithError(err).Error("http: URL %v failed to convert XML to Json resp.Body", *reqURL)
 				} else {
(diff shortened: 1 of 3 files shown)


internal/inputs/http.go
```diff
@@ -89,6 +89,17 @@ func RunHTTP(dataStore *[]interface{}, doLoop *bool, yml *load.Config, api load.
 				if api.Pagination.OriginalURL == "" || (api.Pagination.OriginalURL != "" && resp.StatusCode >= 200 && resp.StatusCode <= 299) && addPage {
 					handleJSON(dataStore, body, &resp, doLoop, reqURL, nextLink)
 				}
+			case contentType == "text/xml" || contentType == "application/xml":
+				jsonBody, err := xj.Convert(resp.Body)
+				if err != nil {
+					load.Logrus.WithFields(logrus.Fields{
+						"err": err,
+					}).Errorf("http: URL %v failed to convert XML to Json resp.Body", *reqURL)
```
ardias: It would be better to use WithError(err) instead of WithFields("err"). It's a bit more "standard".

internal/inputs/http.go
```diff
@@ -112,16 +123,16 @@ func RunHTTP(dataStore *[]interface{}, doLoop *bool, yml *load.Config, api load.
 						}
 						// if it is XML, convert XML to JSON and process it
 					case load.TypeXML:
-						xmlbody := strings.NewReader(strBody)
-						jsonbody, err := xj.Convert(xmlbody)
+						xmlBody := strings.NewReader(strBody)
+						jsonBody, err := xj.Convert(xmlBody)
```
ardias: this is quite a lot of string manipulation, probably not very good memory and cpu wise. We should try to use xml Unmarshall...

Kav91:
@haihongren was there any reason you didn't choose to just unmarshal the xml over converting to json first?

haihongren:
> @haihongren was there any reason you didn't choose to just unmarshal the xml over converting to json first?

@Kav91 @ardias The reason is that we already have handleJSON function in Flex to handle JSON payload very well. I try to avoid introducing additional data structure and redundant code unnecessarily. It was easy enough to convert xml to JSON.

ardias:
You shouldn't need to create another data structure. It should be possible to extract the common code into another func that you call form handleJSON and handleXML. Anyway, since it was already like this, no need to change it right in this commit. We'll change it later

internal/inputs/http.go
Kav91: @ardias I tried to used encoding/xml to do this task, but it cannot whereas @haihongren method does which is why he probably went down this path.

So for whatever reason, xml.Unmarshal does not function the same way as json.Unmarshal does, and cannot unmarshal into an interface{}.

Below example:
- perfectly valid XML from ->  https://www.w3schools.com/xml/note.xml
- enters the correct case for XML
- however when trying to Unmarshal to an interface{} it returns nil, with no error.

image


When using @haihongren approach it works as expected with no issue.

Kav91:
@ardias given the xml.Unmarshal approach is not valid as above, this should be merged as it improves user experience, and also the contentType check will help validate the XML payload earlier. 
Also it is the current processing behaviour when it falls through to the default case anyway, so we are not adding any additional overhead then we already are.

internal/inputs/http.go
ardias: Ok, to be fair I hadn't tried using xml.Unmarshall yet, I assumed it was like Json. Maybe it has different requirements. 
Let's keep it like this for now.

PR #84, opened 2020-02-13
Added unit tests

cristianciutea:
This PR add new unit tests

### DIFF
--- a/internal/processor/flattener.go
+++ b/internal/processor/flattener.go
@@ -7,6 +7,7 @@ package processor
 
 import (
 	"fmt"
+	"github.com/jeremywohl/flatten"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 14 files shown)


internal/processor/flattener.go
```diff
@@ -311,3 +312,55 @@ func contains(s []string, searchterm string) bool {
 	i := sort.SearchStrings(s, searchterm)
 	return i < len(s) && s[i] == searchterm
 }
+
+// RunLazyFlatten lazy flattens the payload
+func RunLazyFlatten(ds *map[string]interface{}, cfg *load.Config, api int) {
```
cristianciutea: Just moved this function

internal/processor/keys_test.go
```diff
@@ -61,49 +62,366 @@ func TestKeyConversionConvertSpace(t *testing.T) {
...
+		keepKeys []string
+		key      string
+		sample   map[string]interface{}
+		expected string
+	}{
+		"WithStrings": {
+			keepKeys: []string{"abc", "xyz"},
+			key:      "def",
+			sample:   getSample(),
+			expected: `{"abc":1,"dfadf":"dafsdfa","xyz":2}`,
```
ardias: shouldn't the expected only be "abc" and "xyz"?

cristianciutea: `RunKeepKeys(keepKeys []string, key *string, currentSample *map[string]interface{}) `

this function will remove the  key *string from the map if is not found in the keepKeys []string

PR #85, opened 2020-02-14
document linux shell command use-case

ardias:
cleaned up some dead code


ardias:
Merging it for now. The docs team will at a later date review when we have the rest of docs ready

PR #86, opened 2020-02-14
Porting the Wiki as traceable MD folders

mariomac:
GitHub wikis are simple but lack on a proper review mechanism to track and review the changes.

It also provides a flat documents index structure, making difficult to group them, making the wiki chaotic as long as it grows.

This PR only copy/pastes the current wiki contents and provides a table of contents.


varas:
wow, thx for this! I guess we can rename the `draft` from the folder name, can't we?

mariomac:
Yes, we can

mariomac:
Done

PR #87, opened 2020-02-16
chore: remove k8s libs

Kav91:
As the infrastructure agent already supports K8s discovery and the option of applying k8s labels these libs and feature are not needed within Flex.

This will also lighten Flex's binary by ~2x.


ardias:
@Kav91 Small question: do you know of anyone using it? Is it "safe" to remove without having to do some kind of public announcement?

Kav91:
@ardias I would say safe, and low impact. I know @hsinghkalsi added it (#68) for a specific customer.

We should take the opportunity to have them move to the supported k8s discovery mechanism rather than continue forward with this since we are listing the discovery feature as deprecated in the docs now. 
They can remain on their current version and move when they are ready.

ardias:
Ok, merging it then.

PR #89, opened 2020-02-17
Reworked Wiki index

mariomac:
Also deprecation notices are written.

### DIFF
--- a/docs/wiki/deprecated/discovery.md
+++ b/docs/wiki/deprecated/discovery.md
@@ -1,7 +1,7 @@
 ### Flex Auto Container Discovery 
 
 > ⚠️ **Notice** ⚠️: this document contains a deprecated functionality that is still
-> provided backwards compatibility. However, we encourage you to
+> provided for backwards compatibility. However, we encourage you to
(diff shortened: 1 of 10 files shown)


docs/wiki/deprecated/discovery.md
```diff
@@ -1,4 +1,9 @@
 ### Flex Auto Container Discovery 
+
+> ⚠️ **Notice** ⚠️: this document contains a deprecated functionality that is still
+> supported by New Relic for backwards compatibility. However, we encourage you to
```
davidbrota: is it supported or not?

in the above section we say: "but
its use is discouraged and unsupported because New Relic provides more convenient implementations
of such functionalities."

mariomac: Rephrased

docs/wiki/deprecated/discovery.md
```diff
@@ -1,7 +1,7 @@
 ### Flex Auto Container Discovery 
 
 > ⚠️ **Notice** ⚠️: this document contains a deprecated functionality that is still
-> supported by New Relic for backwards compatibility. However, we encourage you to
+> provided backwards compatibility. However, we encourage you to
```
davidbrota: "provided for" as in the other replacements

PR #92, opened 2020-02-18
Docs: Adding JMX to the list of unsupported functionalities

mariomac:


ardias:
@mariomac I don't think we are going to deprecate JMX in the near future. NRI-JMX does not allow processing the metrics so JMX in Flex still has some advantages.
Since we don't have full integration tests I would put it in the experimental section

PR #101, opened 2020-02-20
more documentation refatoring

ardias:


ardias:
Also disabled on test that seems to be flaky.

ardias:
Merging this so that the docs team can  start reviewing.
We can only always fix it later

PR #102, opened 2020-02-20
readme cleanup

ardias:
docs: README.md cleanup
docker: small fix to docker build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN mkdir -p /var/db/newrelic-infra/custom-integrations/flexContainerDiscovery/
 # ADD flexConfigs /var/db/newrelic-infra/custom-integrations/flexConfigs/
 
 # copy config/definition/binary over
-COPY ./configs/nri-flex-config.yml /etc/newrelic-infra/integrations.d/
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN mkdir -p /var/db/newrelic-infra/custom-integrations/flexContainerDiscovery/
 # ADD flexConfigs /var/db/newrelic-infra/custom-integrations/flexConfigs/
 
 # copy config/definition/binary over
-COPY ./configs/nri-flex-config.yml /etc/newrelic-infra/integrations.d/
+COPY ./configs/nri-flex-config-linux.yml /etc/newrelic-infra/integrations.d/
 COPY ./configs/nri-flex-def-linux.yml /var/db/newrelic-infra/custom-integrations/nri-flex-definition.yml
```
Kav91: Should this section be updated to use the same paths the agent uses now, eg. /bin/ and /newrelic-integrations/ instead?

ardias: You're right, they should. I'll change that, thanks

ardias: fixed this in another PR @Kav91

PR #109, opened 2020-03-01
feat: read v4 agent integrations config format to ensure config_file arg is fully supported

### LINKED ISSUE #108: Verbose flag has stopped showing topical data
in previous releases of nri-flex, we could easily troubleshoot issues with individual config files by using a snippet like this: 
```
.\nri-flex.exe --verbose --pretty --config_file "path\to\config.yml"
```

This would output the JSON payload that was generated by the config and sent to NRDB, which made it very easy to see issues with formatting. Additionally, the actual execution of the command was listed in this output as well; which could show syntax issues being sent in things like the `run:` function.

Request that this functionality be restored to alleviate frustration, increase UX, and reduce development time for users.

ref: 
image

Kav91:
With users migrating to the GA Version of Flex were we are suggesting to use the new V4 Integrations format, the config_file argument does not understand that agents V4 integrations yaml configuration, so it fails to work properly for users. 

This is where we see many users trying to pass a single config file, and Flex returns that it requires a name for the config as it doesn't understand that format. Which in turn makes it harder to debug single configs.

Fixes #108 . 

cc/ @bpschmitt @haihongren

### DIFF
--- a/cmd/nri-flex/nri-flex_test.go
+++ b/cmd/nri-flex/nri-flex_test.go
@@ -29,6 +29,7 @@ func testSamples(expectedSamples []string, entityMetrics []*metric.Set, t *testi
 			delete(sample.Metrics, "flex.time.startMs")
 			delete(sample.Metrics, "flex.time.endMs")
 			delete(sample.Metrics, "flex.time.elaspedMs")
+			delete(sample.Metrics, "flex.commandTimeMs")
 			out, err := sample.MarshalJSON()
 			if err != nil {
 				load.Logrus.WithFields(logrus.Fields{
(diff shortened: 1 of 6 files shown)


ardias:
Hi @Kav91 I'm not seeing how this is related to the issue you linked above. The user seems to have run flex directly and while yes the error he has in the configuration seems to be related to this issue, I'm not sure we should support this feature.

If user are having trouble moving between flex-only and flex-with-agent, especially regarding the configuration, I think it's more an issue of documentation then adding the new "feature" to flex.

Maybe we should surface this in the initial README instead of just mentioning inside the "wiki" documentation.

bpschmitt:
@ardias so my understanding is that in order to test a Flex config manually, I have to install a separate standalone version of Flex?

Kav91:
@ardias @bpschmitt the issue is the user will need to remove the below portion and re-indent their flex config every time to test manually. Which is a terrible user experience.

eg. deleting this
```yaml
integrations:
  - name: nri-flex
    config:
```

Also it is related to #108 because it is the same version and the only difference if you notice closely in the screenshots is that Zack is executing a flex config from the /integrations.d folder where he is having the issue so it does not run properly because that is not a config Flex understands anymore its an infra agent config that then passes the config section too Flex to run. 

In my opinion this is a huge issue and difference for users. Any config they add into the /integrations.d cannot be manually tested or debugged with the config_file parameter anymore without this PR.

This has come up in slack several times in passing as well for users.

bpschmitt:
@Kav91 @ardias yeah, I have definitely run into this frustration since moving to Infra configs.  I have to maintain 2 different config versions just to test.  Not fun.

ardias:
Can you guys take a look at the current documentation and see if it's clear this separation and how reuse the Flex config files when running with the agent?

https://github.com/newrelic/nri-flex/blob/master/docs/basics/configure.md
https://github.com/newrelic/nri-flex/blob/master/docs/basic-tutorial.md

theletterf:
I think that referencing the flex config files directly from the Infra agent config file would allow you to keep separate Flex config files for testing AND later use them in production without having to move them to the Infra config. 

https://github.com/newrelic/nri-flex/blob/master/docs/basics/configure.md#Linktoaseparateconfigurationfile

ardias:
I'm closing this for now. Let's keep the branch in case we see that it makes sense to open it up again

PR #111, opened 2020-03-04
Updated splitArray to handle map data type

haihongren:
- Updated splitArray to handle map[string]interface{} data type
- Expand nested map as attributes and decorate the array samples


ardias:
@haihongren can you provide an example of the type of data this applies to? What is the difference to SplitObjects? Can you provide documentation for it?

Kav91:
@ardias as the name implies splitting an object vs splitting an array are very different.

SplitArray was added sometime ago, but was not inheriting parent attributes to help distinguish the samples properly @haihongren PR adds that needed functionality. 

We have seen this requirement appear every so often with APIs like megaport, where they supply an array of numerics, rather than an array of objects with a value attribute.

PR #118, opened 2020-03-10
adding confluent kafka flex config

aschmid13:


Kav91:
Nice, this is a pretty sweet example using a fair few capabilities together!

PR #121, opened 2020-03-16
feat: add math_default functionality

Kav91:
Users have encountered APIs that may not consistently return metrics back.
When using math, if a variable is unable to be substituted the attribute will not be created. 
This causes issues with alerting and graphing, as the data becomes inconsistent.

Below functionality has been added to overcome this situation.

### math_default

| Applies to | Description                                                                                     |
| :--------- | :---------------------------------------------------------------------------------------------- |
| API        | When math operations are used, if an attribute is non-existent use the defined default instead. |

```json
{
    "id": "eca0338f4ea31566",
    "leaderInfo": {
        "leader": "8a69d5f6b7814500",
        "startTime": "2014-10-24T13:15:51.186620747-07:00",
        "uptime": "10m59.322358947s",
        "abc": {
            "def": 123
        }
    },
    "name": "node3"
}
```

You could create another attribute that is, for example, the **sum** of attributes `leaderInfo.abc.def` and `leaderInfo.abc.hij`, however in this case `leaderInfo.abc.hij` is not returned in the payload, you could then specify a `math_default` to use instead in your config:

```yaml
name: example
apis:
    - name: someService
      math:
          sum: ${leaderInfo.abc.def} + ${leaderInfo.abc.hij} + 1
      math_default: 1
```

Which would return the following:

```json
"metrics": [{
  "id": "eca0338f4ea31566",
  "leaderInfo.abc.def": 123,
  "leaderInfo.leader": "a8a69d5f6b7814500",
  "leaderInfo.startTime": "a2014-10-24T13:15:51.186620747-07:00",
  "leaderInfo.uptime": 10,
  "name": "node3",
  "sum": 125
}]
```

### DIFF
--- a/internal/load/load.go
+++ b/internal/load/load.go
@@ -291,6 +291,7 @@ type API struct {
 	PercToDecimal    bool              `yaml:"perc_to_decimal"` // will check strings, and perform a trimRight for the %
 	PluckNumbers     bool              `yaml:"pluck_numbers"`   // plucks numbers out of the value
 	Math             map[string]string `yaml:"math"`            // perform match across processed metrics
+	MathDefault      string            `yaml:"math_default"`    // if unable to substitute a math variable use this default
 	SubParse         []Parse           `yaml:"sub_parse"`
 	ValueParser      map[string]string `yaml:"value_parser"`      // find keys with regex, and parse the value with regex
 	ValueTransformer map[string]string `yaml:"value_transformer"` // find key(s) with regex, and modify the value
--- a/docs/basics/functions.md
+++ b/docs/basics/functions.md
@@ -4,13 +4,14 @@ Flex has many useful functions, which can be combined in different ways to help 
 
 -   [Function precedence order](#function-precedence-order)
 -   [Supported functions](#supported-functions)
-    -   [add_attribute](#add_attribute)    
+    -   [add_attribute](#add_attribute)
     -   [convert_space](#convert_space)
     -   [ignore_output](#ignore_output)
     -   [keep_keys](#keep_keys)
     -   [lazy_flatten](#lazy_flatten)
     -   [lookup_file](#lookup_file)
     -   [math](#math)
+    -   [math_default](#math_default)
     -   [perc_to_decimal](#perc_to_decimal)
     -   [remove_keys](#remove_keys)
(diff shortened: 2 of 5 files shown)


internal/load/load.go
```diff
@@ -291,6 +291,7 @@ type API struct {
 	PercToDecimal    bool              `yaml:"perc_to_decimal"` // will check strings, and perform a trimRight for the %
 	PluckNumbers     bool              `yaml:"pluck_numbers"`   // plucks numbers out of the value
 	Math             map[string]string `yaml:"math"`            // perform match across processed metrics
+	MathDefault      string            `yaml:"math_default"`    // if unable to substitute a math variable use this default
```
ardias: @Kav, I think it would make more sense for MathDefault to be a property of Math itself instead. Math would have to be a struct like Parse (SubParse).
Then, config could be like:

    - name: someService
      math:
          sum: ${leaderInfo.abc.def} + ${leaderInfo.abc.hij} + 1
          default: 1

which i think makes more sense. Otherwise, the user may just put math_default anywhere and it wouldn't do anything ant it wouldn't also fail (this is something that happens in a few other cases that we should fix sometime).

Another option could be having the option of defining default values for each of the expressions, a bit like yaml itself supports doing. 

    ${leaderInfo.abc.def:<default_value>}

For example:

    - name: someService
      math:
          sum: ${leaderInfo.abc.def:0} + ${leaderInfo.abc.hij:0} + 1

This way we wouldn't need the **default**

Kav91: @ardias The first suggestion would not support multiple variables, and would be a breaking change.

I think your second idea is the best solution. I'll rework this PR to support the second approach. Thx

theletterf: @Kav91 Since I guess the docs will change following your rework, ping me here once `functions.md` has been edited further and I'll gladly step in to review it.

ardias: @Kav91 the first suggestion would give a default value for the whole "equation" not to the individual values. So if any part of the equation could not be evaluated (or return an error) you would return the default value. The way you have it now, the default seems disconnected from math (apart from the name)

Kav91:
closing PR, will create a new PR with updated functionality based on this feedback.

PR #123, opened 2020-03-16
adding windows examples

thezackm:
adding into `/examples/flexConfigs`
  - windows-file-count-example.yml
  - windows-file-count-lookup.json
  - windows-service-status-example.yml
  - windows-service-status-filtered-example.yml
  - windows-service-status-filtered-lookup.json
  - windows-sftp-connect-example.yml
  - windows-sftp-connect-script.ps1


theletterf:
@ardias Wonder if we should update README and docs to reflect Windows support. When is it going to go GA?

PR #135, opened 2020-04-01
feat: remove keys last

Kav91:
Remove keys should be kept last.

For eg. when wanting to do a math function, and cleaning up prior attributes the keys will be removed and math will fail.

Docs order has been updated and formatted as well.


theletterf:
Docs edits approved. Thanks @Kav91 !

PR #137, opened 2020-04-02
split_array enhancement - added leaf_array option

haihongren:
- When leaf_array is true, split_array turns array leaf node(s) into sample(s)
- Updated functions.md with usage documentation
- Added test for leaf_array feature

### DIFF
--- a/docs/basics/functions.md
+++ b/docs/basics/functions.md
@@ -37,7 +37,7 @@ Flex applies data parsing and transformation functions in a specific order, rega
 2. [start_key](#start_key)
 3. [strip_keys](#strip_keys) *
 4. [lazy_flatten](#lazy_flatten)
-5. [split_array](#split_array)
+5. [split_array (leaf_array)](#split_array)
 6. [split_objects](#split_objects)
 7. Standard flatten (auto-flattening)
 8. [to_lower](#to_lower)
@@ -820,13 +820,14 @@ Which would return the following:
 }]
 ```
 
-### split_array
+### split_array (leaf_array)
 
 | Applies to | Description |
 | :--------- | :---------- |
 | API | Split an array that has nested arrays |
+||With leaf_array option, turn array leaf node(s) into sample(s)
 
-**Example**
+**Example 1 (split_array)**
 
 Consider a service that returns the following payload:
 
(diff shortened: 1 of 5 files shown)


docs/basics/functions.md
```diff
@@ -832,13 +832,14 @@ Which would return the following:
 }]
 ```
 
-### split_array
+### split_array (leaf_array)
```
theletterf, line 835: Does leaf_array need to be there in brackets?

docs/basics/functions.md
```diff
@@ -832,13 +832,14 @@ Which would return the following:
...
 
-| Applies to | Description                           |
-| :--------- | :------------------------------------ |
-| API        | Split an array that has nested arrays |
+| Applies to | Description |
+| :--------- | :---------- |
+| API | Split an array that has nested arrays |
```
theletterf, line 839: s/Split/Splits

@haihongren What does the function split the nested arrays into? New arrays at the root level of the split array?

docs/basics/functions.md
```diff
@@ -832,13 +832,14 @@ Which would return the following:
...
+||With leaf_array option, turn array leaf node(s) into sample(s)
```
theletterf, line 840: Enable the `leaf_array`option to turn array leaf nodes into samples

docs/basics/functions.md
```diff
@@ -917,6 +918,97 @@ Which would return the something like following:
...
+                200,
+                190,
+                180
+            ]
+        }
+    }
+}
+```
+
+You could split the leaf nodes within `timestamps` and `filters` branches into separate samples, then merge and join them together.
```
theletterf, line 947: s/then merge and join them together/which you could then merge and join.

docs/basics/functions.md
```diff
@@ -917,6 +918,97 @@ Which would return the something like following:
...
+    start_key:
+      - concurrent_plays
+      - filters
+    custom_attributes:
+      accountid: "AccountA"
+    merge: myMetricsSample
+    join_key: index
+
+```
+Which would return the something like following:
```
theletterf, line 978: s/return the something like/return the following

docs/basics/functions.md
Kav91: @theletterf given you have a simple array eg. [1,2,3,4] it would split that into 4 new samples.
Eg. sometimes APIs return an array of values, with another array of timestamps in the same payload. 

We need to split both of these out separately so the timestamp with its corresponding value can be maintained properly.

docs/basics/functions.md
haihongren: So user knows there is a leaf_array option with split_array.

docs/basics/functions.md
haihongren: @theletterf  Updated functions.md accordingly.

docs/basics/functions.md
haihongren: Updated functions.md accordingly.

docs/basics/functions.md
haihongren: Updated functions.md accordingly.

docs/basics/functions.md
haihongren: Updated functions.md accordingly.

theletterf:
Good to go @haihongren !

PR #141, opened 2020-04-11
Kav/v4 config file support

Kav91:
Resubmitting #109 
Have also added a warn message if the agent config is tested directly.

cc/ @theletterf jfyi as if this is merged it will improve testing for customers so we should think about updating the docs. I've intentionally not added any docs update in this PR as we can update it separately and resolve some customer concerns quickly.

@jfjoly


theletterf:
@Kav91 Sounds good. Feel free to DM me about the docs for this change.

ardias:
Hi @Kav91 
Can't tell for sure if the PR support loading external files, not just the embedded config format.
In the v4 config format you can also specify and external configuration file:
`integrations:
    name : nri-flex
    config_template_path: /path/to /external/config.yml
`
Is this feature included? I think we should. there a 3rd way to load external files (config folder env var( but we can skip that one for now.

Kav91:
Hi @ardias 
It only supports what is in `config` which is the typical scenario.
In load.go, you'll see the struct defined there, I haven't added support for `config_template_path`.
```
// ConfigEntry holds an integrations YAML configuration entry. It may define multiple types of tasks
type ConfigEntry struct {
	Name   string `yaml:"name"`
	Config Config `yaml:"config"`
}
```

You can update the same branch if you'd like to add the support, or we could add this in a follow up PR.

ardias:
There's a potential issue with supporting `config_template_path` which is that it MUST NOT be a v4 config format file, otherwise we might get into a loop of loading files. It must be a "default" Flex config which some people might already have. Probably not many but doesn't hurt having it and if not might confuse someone if they see an example using it.
We can merge this and I'll try to open a PR to add support for that later

PR #143, opened 2020-04-14
LXC Container example for Ubuntu.

crshanks:


Kav91:
@crshanks nice work!
Re. the dashboard, we could put it in the config file at the bottom and just comment it out.
Or put it in the same folder and call it `lxc-container-ubuntu-dashboard.json`  so they stay together.
What do you think?

crshanks:
Adding example dashboard for lxc-container-ubuntu-example.yml

PR #152, opened 2020-04-24
adias/enable github actions

ardias:
This PR adds a Github Actions workflow file to enable Gihub actions builds.
NR seems to be going in the direction of using Github actions as the CI build engine so instead of moving to travis-ci.com to have windows support, we decided to use github actions.

Some tests are flagged to not run in windows (prometheus) due to some issues with reading the payloads. We fixed the issue in "commands" by having a different set of functions to load the required payloads specifically for windows. We can use the same approach for prometheus later IF we start to officially support it.

This first PR does not output any type of artifacts. It just runs integration tests. 
We will have to add all the rest of the build steps, specifically **release**, at a later stage


.github/workflows/flex.yaml
```diff
@@ -0,0 +1,53 @@
+name: Run tests
+
+on: [push, pull_request]
+# if we want to limit builds on chanes to code only
```
davidbrota, line 4: chanes -> changes

ardias:
Having build on push and pull_request seems to build twice when we push a commit to the PR.
Travis would do the same but the PR build would be done against master (if i'm not mistaken). In this case it does not seem to be the same. It's just two similar builds.

internal/inputs/commands_test.go
```diff
@@ -218,8 +218,9 @@ func TestDf2(t *testing.T) {
 	expected := dataStoreExpected[0].(map[string]interface{})
 	actual := dataStore[0].(map[string]interface{})
 
+	fmt.Println(fmt.Sprintf("%v", actual))
```
carlosroman, line 221:
```suggestion
	t.Logf("%v", actual)
```

ardias: that shouldn't even be there. thanks

PR #159, opened 2020-04-27
test: Added http_test.go to check DNS panic has been fixed

carlosroman:


internal/outputs/http.go
```diff
@@ -44,14 +48,16 @@ func postRequest(url string, key string, data []byte) error {
...
-	defer resp.Body.Close()
+
+	defer func(body io.ReadCloser) {
+		_, _ = ioutil.ReadAll(body)
+		_ = body.Close()
+	}(resp.Body)
 
 	if resp.StatusCode > 299 || resp.StatusCode < 200 {
-		err = fmt.Errorf("http: post failed, status code: %d", resp.StatusCode)
+		return fmt.Errorf("http: post failed, status code: %d", resp.StatusCode)
```
ardias, line 60: why this change?

internal/outputs/http.go
```diff
@@ -44,14 +48,16 @@ func postRequest(url string, key string, data []byte) error {
...
 		load.Logrus.WithError(err).Error("http: failed to send")
+		return errors.Wrap(err, "http: failed to send")
 	}
-	if resp == nil {
-		return fmt.Errorf("http: response nil")
-	}
-	defer resp.Body.Close()
+
+	defer func(body io.ReadCloser) {
+		_, _ = ioutil.ReadAll(body)
```
ardias, line 55: do we need to read the whole body before closing?

internal/outputs/http.go
carlosroman: Because it basically did that two lines down after the `if` statement. Felt it was more obvious this was the return value here as we never process `err` after this.

internal/outputs/http.go
carlosroman: It is good practice to read the all the body of a http request made in Go.
>Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.
https://golang.org/pkg/net/http/#Client.Do

Just making sure we tidy up resources we've used.

PR #162, opened 2020-04-28
update command API with some windows info

ardias:


docs/apis/commands.md
```diff
@@ -104,6 +120,26 @@ apis:
         split_by: \s+
 ```
 
+As stated before, in Windows, you can also override the default shell, `cmd`, with, for example, powershell.
```
thezackm, line 123: update to add code format around "powershell": 
```
for example, `powershell`.
```

docs/apis/commands.md
```diff
@@ -104,6 +120,26 @@ apis:
...
+  - name: windowsServiceList
+    commands:
+      - run: service
+        shell: powershell
+        event_type: WindowsCommand
+        split: horizontal
+        set_header: [status,name,displayname]
+        row_start: 1
+        regex_match: true
+        split_by: (\w+)\s+(\w+)\s+(\w+)
```
thezackm, line 138: would highly recommend updating this to a more "PowerShell Native" option; `service` is an Alias for `Get-Service` and there's also a risk that the default output could get messy on longer service names that are automatically truncated in the console.

ex: 
image


recommended update: 
`- run: Get-Service | Format-Table -Property Status, Name, DisplayName -Autosize`

image


alternatively, it might be beneficial to show how [conversion to JSON](https://github.com/newrelic/nri-flex/blob/master/examples/flexConfigs/windows-service-status-example.yml) will remove the need for the majority of this example

docs/apis/commands.md
```diff
@@ -104,6 +120,26 @@ apis:
...
+```
+
+In this example we are executing a command, `service`, using Powershell as the command shell.
```
thezackm, line 141: replace: `Powershell` with `PowerShell`

if the above update is accepted; this needs to be changed as well:

> In this example we are executing the command `Get-Service`, using PowerShell as the command shell and ensuring proper spacing in the output table with the `-Autosize` parameter

docs/apis/commands.md
ardias: I changed the command to the native option. But looking at the example of converting to Json, to be honest it seems a lot easier to process it this way than to use powershell to output in json. Using Flex to process the data also emphasizes the "power" of Flex, which is the simplicity. That powershell example is a "beast" and not many people will ever do that kind of thing.

docs/apis/commands.md
ardias: I've changed all to lowercase since that was the most common.

thezackm: `PowerShell` is the proper case for the shell, we use lowercase in the Flex `shell` selection, but anywhere else it should be properly formatted IMO

docs/apis/commands.md
thezackm: I'm ok with that; "beast mode" can be left in the docs

docs/apis/commands.md
ardias: ok, makes sense

PR #163, opened 2020-04-29
adias/windows http test

ardias:


integration-test/https-server/server.go
```diff
@@ -0,0 +1,115 @@
...
+			 "memory": 3500,
+			 "disk": 500
+			} 
+		]
+	}
+	`))
+}
+
+func getCertificates(certFile string, keyFile string) (certFilePath string, keyFilePath string) {
+	if runtime.GOOS == "windows" {
```
ardias, line 107: this is because in Linux we are running in a docker container and the cabundle dir is in the root alongside the executables, but in windows we are running the server file with "go run" which changes the working dir

Noly: Can you add a comment explaining this?

json
```diff
@@ -0,0 +1,11 @@
+
```
varas, line 1: I guess this isn't being used

integration-test/https-server/server.go
```diff
@@ -0,0 +1,118 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	certFile = "cabundle/cert.pem"
```
carlosroman, line 15: Just a minor suggestion:
```suggestion
	certFile = path.Join("cabundle", "cert.pem")
```
If you do add this `path` will need to be added to the imports + `keyFile` should be updated as well.

integration-test/https-server/server.go
```diff
@@ -0,0 +1,118 @@
...
+		MinVersion:               tls.VersionTLS12,
+		CurvePreferences:         []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256},
+		PreferServerCipherSuites: true,
+		CipherSuites: []uint16{
+			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
+			tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+		},
+		InsecureSkipVerify: true,
```
varas, line 66: isn't this a client option?

integration-test/https-server/server.go
```diff
@@ -0,0 +1,118 @@
...
+
+	err := srv.ListenAndServe()
+	if err != http.ErrServerClosed {
+		logrus.WithError(err).Error("failed to start http server")
+	}
+}
+
+func startHTTPS(mux *http.ServeMux) {
+	cfg := &tls.Config{
+		MinVersion:               tls.VersionTLS12,
```
varas, line 57: why not default config?

integration-test/https-server/server.go
ardias: Yeah i think it is. I probably left it when I was doing some tests.

integration-test/https-server/server.go
ardias: this was already here, so I didn't change it

json
ardias: oops, yeap not used

PR #164, opened 2020-04-29
file documentation

davidbrota:


docs/apis/file.md
```diff
@@ -0,0 +1,73 @@
...
+* [Basic usage](#Basicusage)
+* [Configuration properties](#Configurationproperties)
+* [Advanced usage](#Advancedusage)
+
+##  <a name='Basicusage'></a>Basic usage
+
+```yaml
+name: example
+apis:
+  - name: linuxDirectorySize
```
thezackm, line 14: is this the right value to use for `name` here? this seems like a copy/paste mistake

docs/apis/file.md
```diff
@@ -0,0 +1,73 @@
...
+    "startTime": "2014-10-24T13:15:50.072007085-07:00",
+    "state": "StateFollower"
+}
+``` 
+
+The generated sample will contain the following attributes:
+
+```
+"leader": "8a69d5f6b7814500",
+"timestamp": "2014-10-24T13:15:51.186620747-07:00",
```
thezackm, line 70: should there be a note here about sending a non-Unix timestamp to NRDB? this would be treated as a string on ingest right?

docs/apis/file.md
```diff
@@ -0,0 +1,73 @@
...
+"uptime": "10m59.322358947s"
```
thezackm, line 71: similar to above; this would be a string in NRDB; I think for external users it can be misleading as they would more than likely expect a float for this value. Is it possible to use a different payload for this example to avoid the confusion of data types?

docs/apis/file.md
davidbrota: yes, good catch

docs/apis/file.md
davidbrota: OK, I'll change it to a unix timestamp. Keep in mind that the same payload example appears also on the functions documentation. https://github.com/newrelic/nri-flex/blob/master/docs/basics/functions.md

thezackm: @theletterf ^^^ we might need to address the `functions.md` file referenced here to avoid confusion with sending the wrong data type in a payload

PR #165, opened 2020-04-30
ci: Adding support for arm and arm64

carlosroman:
Updated goreleaser config to build `arm` and `arm64` Linux versions.


ardias:
why?

carlosroman:
> why arm?

Adding support for running Flex on things like the Raspberry PI and Amazon's Arm-Based EC2 Instances.

PR #167, opened 2020-05-02
chore/feat: allow command output to be cached, update test & docs

Kav91:
* add cache to docs
* allow command output to be cached (allows post processing on raw output without re-executing the same commands)
* update tests
* rename **yml** to **cfg** in commands.go for better readability


ardias:
@Kav91 Can this maybe create some confusion with the other `cache` directive?
Also, from what I'm seeing  it's similar to `store_variables` but for a whole "dataset", correct?

Kav91:
@ardias quite different it stores raw output, so you can post process commands multiple times over it. With store variable it can only happen when metric sets are being created and substituted at the beginning. That would then be a subsequent api block, so its not designed to be used within the same api block and commands or handle multi line or large outputs.

Regardless of all that, I think best to hold off on this PR anyway, it was for a customer use case that came up where they wanted to run commands over the same data set in different scenarios without fetching it again. But this particular functionality doesn't seem to be a common requirement.

For context some of the scenarios were:
-  a set of metrics are returned on each line, process those metrics, but also report how many rows or items were in this. There is no unique identifier in this scenario to do a unique count of the samples.
- you get a json payload where it has multiple arrays, process that data, but also get the length of those arrays and report those lengths in another sample.

Lets give it till next week sometime before closing, while I collect feedback.
Also open to suggestions in solving those scenarios mentioned as well.

ardias:
thanks @Kav91  
Regarding those scenarios you mention I almost think it would be possible to support them by adding "diagnostic" data while processing them, similar to what is done with the configs (number of config processed, etc.. that is returned in flexStatusSample). 
It could be enabled by an option to return the data or not. The case of storing that data in another sample may be trickier. Perhaps having another operator "select into" like sql? :)

ardias:
Closing this for now as it is not required at the moment. If we see value in it later we may re-open it.

PR #168, opened 2020-05-04
add windows file integration test

davidbrota:


integration-test/file_integration_windows_test.go
```diff
@@ -0,0 +1,37 @@
...
+	tests := map[string]struct {
+		configFile string
+		want       Metrics
+	}{
+		"jsonEtcdSelf": {"configs/windows-json-test.yml", wantJsonMetrics},
+		"csvTest":      {"configs/windows-csv-test.yml", wantCsvMetrics},
+	}
+
+	// copy payloads to the current user's TMP folder plus a folder with spaces
+	assert.NoError(t, exec.Command("cmd", "/C", "xcopy", "/y", "..\\test\\payloads\\etcdSelf.json", filepath.Join("%TMP%", "test payloads", "etcdSelf.json*")).Run())
```
carlosroman, line 25: It might be better to use `ioutil.TempDir`. That way you can also get it to tidy up after the test runs.
A good example of this is https://github.com/newrelic/nri-flex/blob/master/internal/outputs/file_test.go#L33.

PR #169, opened 2020-05-04
fix: rename_keys respects use of regex substitutions - saves a step 

sschwartzman:


Kav91:
LGTM - needed update/fix.

PR #170, opened 2020-05-05
chore: add valueMapper test

Kav91:


Kav91:
merging as minor adjustment & adding test since it is an existing undocumented feature

PR #174, opened 2020-05-07
Zack docs updates

thezackm:
updating the following: 

`lookup_file` >> based on user feedback re: confusion on unused properties in the lookup json example, what the service payload was meant to represent, and multiple instances of "name"

`timeout` >> changed from 10 to 10000 and added note about ms based on [default setting](https://github.com/newrelic/nri-flex/blob/ae8a6f6d4626884a84715988a61679679a0c3be7/internal/load/load.go#L106)


theletterf:
Will check this soon!

PR #180, opened 2020-05-20
http headers on sample

Noly:
Url API now can retrieve response header value attached on each generated sample included a response error.
Setting up `return_headers` to true


internal/inputs/http_test.go
```diff
@@ -7,102 +7,318 @@ package inputs
...
+						URL:           "/todos",
+						Timeout:       5100,
+						ReturnHeaders: true,
+					},
+				},
+			},
+			[]interface{}{
+				map[string]interface{}{
+					"error":                     "Missing Required Parameters",
+					"api.StatusCode":            200,
```
ardias, line 217: this is strange. If statusCode is 200, "error" should not be present. I think error is only when the http call fails, no?

ardias:
Looks good. Check the windows test please since they seem to be failing

internal/inputs/http_test.go
Noly: Yes, totally. As an expected value could be more real with a 5xx status code. Will change it.

However I've saw apis that return success status code with error message in body XD

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## 1.3.1 - 2020-05-XX
+### Added
+- Url API now can retrieve response header value attached on each generated sample included a response error.
```
ardias, line 10: I would rephrase as :
Url API can now include all the response headers in the resulting sample. Set `return_headers` to true to enable.

docs/apis/url.md
```diff
@@ -105,4 +106,18 @@ apis:
             keys: [net.connectionsReading, net.connectionsWriting, net.connectionsWaiting]
     math:
       net.connectionsDroppedPerSecond: ${net.connectionsAcceptedPerSecond} - ${net.handledPerSecond}
-```
\ No newline at end of file
+```
+
+##  <a name='ReturnResponseHeaders'></a>Include response headers on sample
+
+To include response headers on the metric sample setting `return_headers` attribute to true is needed.
```
ardias, line 113: "To include response headers on the metric sample set `return_headers` attribute to true."

internal/inputs/http.go
```diff
@@ -246,50 +246,40 @@ func setRequestOptions(request *gorequest.SuperAgent, yml load.Config, api load.
...
-			// load.StoreAppend(theSample)
-			*dataStore = append(*dataStore, theSample)
+	var cb responseBody
 
-			if theSample["error"] != nil && fmt.Sprintf("%v", theSample["error"]) != "false" {
-				load.Logrus.Debugf("http: request failed %v", theSample["error"])
-			}
+	switch t := b.(type) {
+	case []interface{}:
+		cb = newObjectBody(t)
```
ardias, line 262: shouldn't this be the other way around? In this case the type `t` is a array on "objects" but you are using 
`newObjectBody` instead of `newArrayBody()`?

ardias:
I'm not completely sure if the new code is equivalent as the old one.

internal/inputs/http.go
Kav91: looks like the functions just needs to be renamed?

CHANGELOG.md
Noly: 👍

docs/apis/url.md
Noly: 👍

internal/inputs/http.go
Noly: Yes, indeed. Thanks

Noly:
> I'm not completely sure if the new code is equivalent as the old one. We may need more robust integration tests for HTTP

Refactor was made after tests were green, manual testing were success. 

Which use case would you like to add as integration test to be more confident with this changes?

ardias:
Tested with **examples/flexConfigs/http-json-example.yml** and the output is the same as the latest release, so I think we're good

PR #181, opened 2020-05-22
Fix numbers to string conversion for lookup files

brushknight:


internal/config/substitutions_test.go
```diff
@@ -68,3 +69,47 @@ func TestSubTimestamps(t *testing.T) {
...
+}
+
+func Test_getSprintfTemplateForVariable(t *testing.T) {
+	var valueInt int = 2456853
+	assert.Equal(t, "%d", getSprintfTemplateForVariable(valueInt))
+	var valueFloat32 float32 = 2456853
+	assert.Equal(t, "%f", getSprintfTemplateForVariable(valueFloat32))
+	var valueFloat64 float64 = 2456853
+	assert.Equal(t, "%f", getSprintfTemplateForVariable(valueFloat64))
+	var valueString string = "2456853"
```
Noly, line 111: You can use also use short variable declaration:
       `valueString := "2456853"
         valueInt := 2456853
         valueFloat64 := float64(2456853)
         valueFloat32 := float32(2456853)` 

Go will infer the data type when is string or int. For float you can use the type builtin

brushknight: Yeah, I was thinking that way, but then turned it into explicit type declaration. Reason is that it will be more visible
- int and we expect %d, 
- string and %s, etc. 
Not sure is it worth it or not.

internal/config/substitutions.go
```diff
@@ -45,39 +45,67 @@ func SubLookupFileData(configs *[]load.Config, config load.Config) error {
...
-			load.Logrus.Debug("config: lookup file needs to contain an array of objects")
+			return &newCfg, nil
 		}
+	default:
+		load.Logrus.Debug("config: lookup file needs to contain an array of objects")
+	}
+	return nil, nil
+}
+
+func getSprintfTemplateForVariable(value interface{}) string {
```
ardias, line 98: there's no need to have such a specific name for the function. maybe a simple "toString(..)" or "formatToString(...)".
also perhaps instead of returning the format string, why not return the formatted value already?

PR #184, opened 2020-05-27
Create linux-file-list.yml

lchapman4:
Adding a new example file that could be used to verify if files are created in a folder.


examples/flexConfigs/linux-file-list.yml
```diff
@@ -0,0 +1,17 @@
+# This integration returns a list of files in a particular directory. 
+# In the past, this integration was used by a customer to verify if files arrived successfully onto a SFTP file share.
```
ardias, line 2: I wouldn't  include "used by a customer".

lchapman4: Done @ardias .  Let me know if I missed anything else.

Issue #178, opened 2020-05-15
experimental command declarations 


jcoope02:
currently the examples given that have experimental features or commands do not declare that. For example the port test example (https://github.com/newrelic/nri-flex/blob/master/examples/flexConfigs/port-test-example.yml) uses the -dial command. If I am missing the dependencies it will not work and will never know why and end up wasting time trouble shooting. On top of that it if you go to the command docs (https://github.com/newrelic/nri-flex/blob/master/docs/apis/commands.md) -dial is not there. Would nice to have it there, but listed as experimental so that I can know how it works, what it does and the dependencies. 

In any case people should know in advance if they are using some thing that is not GA

theletterf:
@ardias @Kav91 @thezackm We should either add a comment to examples using experimental features, or move them to an `experimental` subfolder.

thezackm:
i think a comment makes the most sense; assumption is there are examples that have both supported and experimental functions in play right now. can we get a mtg scheduled to outline the reorg of the examples directory? I can make myself available to fit schedules; just toss up some times against my calendar and let's kick this bucket!

thezackm:
resolved with PR #185

PR #185, opened 2020-05-27
marking examples with experimental functions

thezackm:
adding comments to all example `.yml` files that have functions from the [experimental](https://github.com/newrelic/nri-flex/tree/master/docs/experimental) docs. this is a fix for Issue #178

### DIFF
--- a/examples/flexConfigs/postgres-db-example.yml
+++ b/examples/flexConfigs/postgres-db-example.yml
@@ -1,4 +1,4 @@
-# NOTE: 'database' is an experimental function at this time
+# NOTE: 'database' is an experimental API at this time
 # ref: https://github.com/newrelic/nri-flex/blob/master/docs/experimental/db.md
 ---
 integrations:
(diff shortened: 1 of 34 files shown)


examples/flexConfigs/postgres-db-example.yml
```diff
@@ -1,3 +1,5 @@
+# NOTE: 'database' is an experimental function at this time
```
ardias, line 1: ' database' is more of an API thatn function. I'd just change that

examples/flexConfigs/sqlserver-database-example.yml
```diff
@@ -1,3 +1,5 @@
+# NOTE: 'database' is an experimental function at this time
```
ardias, line 1: ' function' - > ' API'

examples/flexConfigs/vertica-database-example.yml
```diff
@@ -1,3 +1,5 @@
+# NOTE: 'database' is an experimental function at this time
```
ardias, line 1: `function' -> 'API'

PR #194, opened 2020-06-17
Fix docs links (which were missing /docs)

njvrzm:
These could also be fixed by removing the leading slash, but I'm not sure which is preferable.


Kav91:
Thanks! merging now as some users have hit the broken links during the new enablement content.

PR #196, opened 2020-06-22
Add a mention of JSON to the command API doc

jfjoly:
JSON format is supported by the command API but there is no mention of it. 
This is an example using it https://github.com/newrelic/nri-flex/blob/master/examples/flexConfigs/osquery-example.yml


theletterf:
Approved. Thanks @jfjoly !

PR #207, opened 2020-08-03
Linux process discovery default to false

Kav91:
This capability is not commonly used anymore.

Disabling it to false will decrease the size of flexStatusSample (important for new pricing model) and provide a performance increase for linux users.

We can consider deprecating the functionality as a second step later.

cc/ @thezackm @ardias


Kav91:
Merging as discussed with José.

PR #212, opened 2020-08-15
feat: implement jq parsing

Kav91:
Provide easy parsing with jq. 

- Other then just having jq's powerful parsing available, due to the design of Flex this is also compatible with csv/xml/command based payloads as they are transformed to interfaces internally before the `gojq` package begins processing.
- As this is implemented at the beginning of the data processing, the execution of other Flex transformation functions will happen after as usual so all prior behaviour is unchanged.

cc/ @haihongren @bpschmitt @thezackm 

#### How it works

```json
{
  "id": "eca0338f4ea31566",
  "leaderInfo": {
    "leader": "8a69d5f6b7814500",
    "startTime": "2014-10-24T13:15:51.186620747-07:00",
    "uptime": "10m59.322358947s",
    "abc": {
      "def": 123,
      "hij": 234
    }
  },
  "name": "node3"
}
```

To easily access and return just the data under `abc` you can use the following configuration:

```yaml
name: example
apis:
  - name: someService
    jq: ".leaderInfo.abc" #<---
```

Returns:

```json
{
  "def": 123,
  "hij": 234
}
```


internal/processor/data.go
```diff
@@ -60,3 +69,44 @@ func processDataSet(dataSet *map[string]interface{}, samplesToMerge *load.Sample
 		CreateMetricSets(mergedData, cfg, i, true, samplesToMerge, originalAPINo)
 	}
 }
+
+func runJq(dataSets interface{}, api load.API) []interface{} {
+	// due to the current design, dataSets is always a slice
+	// so an api that returns a map will still be a slice returned for processing
+	// for better usability if there is no array prefix access the first item in the slice automatically for the user
+	if !strings.HasPrefix(api.Jq, ".[") {
```
haihongren, line 77: **Additional scenario:** 
If it is a simple `api.Jq: '.'`, we do not need to add `".[0]"`.
Although it does not do much using `api.Jq: '.'`, it was the first thing i did for the test.

internal/processor/data.go
```diff
@@ -60,3 +69,44 @@ func processDataSet(dataSet *map[string]interface{}, samplesToMerge *load.Sample
...
+		case []interface{}:
+			return value
+		case map[string]interface{}:
+			return []interface{}{value}
+		case error:
+			load.Logrus.WithFields(logrus.Fields{
+				"api": api.Name,
+				"jq":  api.Jq,
+			}).WithError(err).Debug("jq: failed to process")
+			return []interface{}{}
```
haihongren, line 106: @Kav91 
- I always have this dilemma: 

1. do we return the original dataSets when error
2. do we return empty data

Kav91: @haihongren was experiencing the same dilemma.
I felt that if someone was using the jq functionality, and it didn't do its job we shouldn't return the dataset. 
So at least they can go look at the error and realize it was because of that. Otherwise if we return and they do subsequent transformations the output would be wrong anyway.

internal/processor/data.go
Kav91: resolving as discussed

internal/processor/data.go
haihongren: @kav91 Agreed.  Nice job adding built-in `jq` support in Flex!

PR #217, opened 2020-08-27
Adding xml format support for commands function

haihongren:
**Adding  xml format support for `commands` function.** 

The file content must have the following signature:

```
<?xml version="1.0"
```

Sample XML file: 
```
<?xml version="1.0" encoding="UTF-8"?>
<catalog>
   <book id="bk101">
      <author>Gambardella, Matthew</author>
      <title>XML Developer's Guide</title>
      <genre>Computer</genre>
      <price>44.95</price>
      <publish_date>2000-10-01</publish_date>
      <description>An in-depth look at creating applications with XML.</description>
   </book>
   <book id="bk102">
      <author>Ralls, Kim</author>
      <title>Midnight Rain</title>
      <genre>Fantasy</genre>
      <price>5.95</price>
      <publish_date>2000-12-16</publish_date>
      <description>A former architect battles corporate zombies, an evil sorceress, and her own childhood to become queen of the world.</description>
   </book>
</catalog>
```

Sample Flex integration file: 

```
integrations:
  - name: nri-flex
    interval: 30s
    config:
      name: xmlFileExample
      apis:
        - name: xmlFileExample
          commands:
            - run: cat myexamples/simpleXML.xml
              output: xml
```

### DIFF
--- a/internal/inputs/commands.go
+++ b/internal/inputs/commands.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 	"time"
 
-	xj "github.com/basgys/goxml2json"
 	"github.com/newrelic/nri-flex/internal/formatter"
 	"github.com/newrelic/nri-flex/internal/load"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


Kav91:
@haihongren nice one!

One suggestion, in the past I had an issue with the signature check for HTTP as not all XML payloads contain that.

In the command struct there is an Output attribute, that we could have accept "xml", so if the user sets that, we can then forceably attempt xml conversion.

I think this way we have a solution for all use cases. 

eg. if `output: xml`

```
xmlBody := strings.NewReader(strBody)
body, err := xj.Convert(xmlBody)

if err != nil {
// fail
}else{
// process
}
```

Kav91 (approved):

haihongren:
@Kav91 
Great suggestion!  
I have updated the code to have ` || commandOutput == load.TypeXML ` logic. 
It will support both `output: xml` and the signature.

Issue #206, opened 2020-08-02
HTTP request handling broken during refactor


Kav91:
Internal discussion: https://newrelic.slack.com/archives/CPVCA6Y86/p1596379457402500?thread_ts=1596373059.401200&cid=CPVCA6Y86

Automatic pagination walk is no longer working as expected which breaks how normal http functionality should work as there are missing payloads from the initial request.  
eg. looks like it ignores the initial results, and are not appended and you just get the last set.

I reverted to my prior http code which was before GA, and everything worked as expected.  Haven't figured out which changes caused this issue.

Kav91:
@Noly are you able to have a look plz? As #180 was the biggest change to how the http functionality worked recently. I'm assuming its related but could be wrong.

Also would be good if we can get on this early as Mike has a live test case that can be used to validate against.

Noly:
Hi @Kav91 thanks for raising this up. We added it on our queue and we will working on this ASAP

Noly:
A new version of Flex [v1.3.5](https://github.com/newrelic/nri-flex/releases/tag/v1.3.5) was released with the fix for this issue.

Thanks for your feedback.

PR #221, opened 2020-09-09
fix http api dropping data from pagination

Noly:
Fix for the regression issue #206 

---

Http API manage calls for responses with pagination in order to collect all data.


carlosroman (changes requested):
Nice catch 👍

carlosroman (approved):

internal/inputs/http.go
```diff
@@ -279,7 +279,7 @@ func handleJSON(sample *[]interface{}, body []byte, resp *gorequest.Response, do
 		*doLoop = false
 	}
 
-	*sample = ds.build()
+	*sample = append(*sample, ds.build()...)
```
varas, line 282: 👏

varas (approved):

PR #222, opened 2020-09-10
twitter-monitor.yml example

CharlieWinters:
Twitter integration from the flex webinar: https://newrelic.com/resources/webinars/your-observability-data-in-one-place1


examples/twitter-monitor.yml
```diff
@@ -0,0 +1,47 @@
+# This integration does a post request to the twitter auth api, uses a token from the auth api to query the tweet api.
+# As an added benefit it then does a third query to analyses each tweets sentiment.
+# Check out https://newrelic.com/resources/webinars/your-observability-data-in-one-place1 for live demo.
+integrations:
+  - name: nri-flex
+    interval: 60m
+    config:
+      name: twitterInfo
+      #point lookup file at a json file that contains api key and api secret key
+      lookup_file: /home/ubuntu/apikeys.json
```
thezackm, line 10: please add a sample lookup JSON file to this PR to be submitted with this config.
ref: [example](https://github.com/newrelic/nri-flex/blob/master/examples/openssl-expiration-lookup.json)

thezackm (changes requested):

thezackm (changes requested):
looks solid, please rename to `twitter-monitor-lookup.json` and we should be ready to rock!

CharlieWinters:
Thanks @thezackm, changed now, also changed the name in the  twitter-monitor.yml to match.

thezackm (approved):
let's GTM

PR #225, opened 2020-09-18
Added Windows uptime samples using PowerShell 5 and PowerShell Core 6+.

sdewitt-newrelic:


thezackm (commented):
question: would it make sense to add an `interval` attribute to these as a commented-out option? it might be a good idea to explicitly show the option for data like this that might not have a need for 30s granularity. 

another idea: adding an additional attribute to these samples for a unix timestamp of last boot time

neither of these are required; just commenting ideas. lmk what you think and I can finish the review.

sdewitt-newrelic:
@thezackm Updated the configs to have the example/commented out interval settings.

thezackm (approved):
let's GTM

PR #227, opened 2020-09-30
Plex Media Server Session Monitoring

icpenguins:
This is an example for using Flex to monitor Plex media server sessions.


thezackm (approved):
let's GTM

thezackm (changes requested):
please change the file location; should be in the root of `/examples`, not in the `examples/flexConfigs` sub-directory

icpenguins:
@thezackm sorry about that! Don't know how I did that in the first place.

thezackm (approved):
thanks @icpenguins!

PR #229, opened 2020-10-06
Update mysql-database-example.yml

lchapman4:
Added a more verbose mysql example with additional monitoring queries and more verbose authentication string with an aws rds example.


Kav91 (approved):
awesome lgtm 👍

PR #237, opened 2020-10-23
Add example for Elastic APM Integration

ericmittelhammer:
Queries Elasticsearch for `Transaction`, `Metric`, and `Error` events collected by Elastic APM


thezackm (changes requested):
please make the following edits:

* move files from `examples/elastic-apm/*` to `examples/*`
* The event limit is set to 500 by default, but it is a [configurable option](https://github.com/newrelic/nri-flex/blob/af2f8cb41fa9f221b2bc1768ecb2698902176d05/configs/nri-flex-config-linux.yml#L21) in the `env:` attribute; please add the option, commented out, to the samples; or a note in the top comments if you think it would be a common problem

ericmittelhammer:
@thezackm: addressed those changes.

thezackm (approved):

PR #238, opened 2020-10-23
added snowflake-example.yml for capturing snowflake table and warehou…

danifitz:
…se metrics

needed to integrate with snowflake so added some examples of how to use snowsql to pull performance data from the internal schema's and send it to New Relic.


danifitz:
license/snyk and security/snyk checks are failing - when i click on details I get an error so I can't check why. What do I need to do to make these checks pass?

thezackm (approved):

PR #240, opened 2020-10-28
Replaced err with pingError in debug logging

### LINKED ISSUE #239: "database: ping error" but no error code
I'm working with a customer on retrieving data from a PostgreSQL server. We're currently getting the following logs:

```
INFO[0000] com.newrelic.nri-flex                         GOARCH=amd64 GOOS=linux version=1.3.2
DEBU[0000] config: git sync configuration not set
WARN[0000] config: testing agent config, agent features will not be available
DEBU[0000] config: running async                         name=VPNTTG
DEBU[0000] config: processing apis                       apis=1 name=VPNTTG
DEBU[0000] fetch: collect data                           name=VPNTTG
DEBU[0000] database: process queries                     database=postgres name=VPNTTG
DEBU[0000] database: db.Ping finished
DEBU[0000] database: ping error                          database=postgres err="<nil>" name=VPNTTG
DEBU[0000] processor-data: running data handler          name=VPNTTG
DEBU[0000] config: finished variable processing apis     apis=1 name=VPNTTG
INFO[0000] flex: completed processing configs            configs=1
```

Notice the `DEBU[0000] database: ping error                          database=postgres err="<nil>" name=VPNTTG` message.

This triggered me to jump into the code to see what's happening and I noticed the following line.

https://github.com/newrelic/nri-flex/blob/a5585dc953d99864d2d8cc2f0496b6189dad1b7d/internal/inputs/database.go#L66

This line is in a `if pingError != nil {` section, so I was wondering if we should replace 'err' with 'pingError' or add 'pingError' to the logging message. 

The goal is to improve the debug output as I currently have no idea why the flex config is failing, though it is probably a database connection error.

kidk:
The current output (as observed at a customer) is nil. This is because `if err != nil` is done a couple of lines higher, so this code wouldn't run if err != nil. Replacing err with pingError should return the right information for the customer to debug the error code.

This fixes #239

### DIFF
--- a/internal/inputs/database.go
+++ b/internal/inputs/database.go
@@ -63,7 +63,7 @@ func ProcessQueries(dataStore *[]interface{}, yml *load.Config, apiNo int) {
 
 	if pingError != nil {
 		load.Logrus.WithFields(logrus.Fields{
-			"err":      err,
+			"err":      pingError,
 			"name":     yml.Name,
 			"database": api.Database,
 		}).Debug("database: ping error")


Kav91 (approved):

Kav91:
Good catch, thanks!

PR #241, opened 2020-11-06
Add GCP as a serverless mode to Flex

msummers-nr:
- Refactored `internal/integration` as `internal/runtime` and added support for running Flex as a GCP Function
- Refactored code tested in AWS as a Lambda, GCP as a Function, and standalone ('default')
- All tests pass

@hsinghkalsi @sschwartzman @prodjito


Kav91 (approved):
Wow nice work!
@haihongren can you run your eyes over this one as well plz? 
LGTM, but its a chunky one so not sure if I missed anything.

function.go
```diff
@@ -0,0 +1,60 @@
...
+	run()
+	log.Debugf("FlexHTTP: exit")
+}
+
+type PubSubMessage struct {
+	Data []byte `json:"data"`
+}
+
+// Target Function endpoint if you want to use Pub/Sub  with the Cloud Scheduler
+func FlexPubSub(ctx context.Context, m PubSubMessage) error {
```
haihongren, line 34: @msummers-nr 
Nice Job ! 
One question, I might have missed it somewhere, just wondering how the PubSubMessage `m` data is passed to Flex for processing.

msummers-nr: It's not, I should add that as a comment. The Pub/Sub endpoint is used by the Cloud Scheduler to run Flex on a periodic basis- it's the Flex "heartbeat"

msummers-nr:
@haihongren @Kav91 I merged the latest changes in without error and Travis failed but I'm not seeing an error in the log. Do you see anything?

function.go
haihongren: @msummers-nr 
Got you!  Thanks for the explanation. 

Just a note, Flex could potentially process the message data from Pub/Sub and inject the data into NR. Currently, Flex running as a Lambda takes the Lambda event payload and processes it using `flex-lambda-ingest.yml` if it exists.   

Please see the Lambda implementation below: 
https://github.com/newrelic/nri-flex/blob/9a972059d0cb485bc1548a4d47e11d25ed02e838/internal/runtime/lambda.go#L76

PR #242, opened 2020-11-10
added flex config to capture Microsoft Azure Application Insights metrics

harrykimpel:


thezackm (commented):
@harrykimpel - would it make sense to use a [base_url](https://github.com/newrelic/nri-flex/blob/master/docs/apis/url.md#SpecifyacommonbaseURL) in this config for easier readability? 

also; what is the purpose of the lookup file in this example? are there situations where you would run a single flex config against `n+1` apps?

Kav91 (approved):
Awesome work!

harrykimpel:
@Kav91 @thezackm great suggestions with base_url and the variable_store ... I will have a look into it. do you want me to update the file and create another PR?

harrykimpel:
> @harrykimpel - would it make sense to use a [base_url](https://github.com/newrelic/nri-flex/blob/master/docs/apis/url.md#SpecifyacommonbaseURL) in this config for easier readability?
> 
> also; what is the purpose of the lookup file in this example? are there situations where you would run a single flex config against `n+1` apps?

the purpose of the lookup file was just to split the yml from any potential IDs/keys

harrykimpel:
@thezackm regarding the failed tests, is this something for me to fix? I do not seem to have access to Snyk

harrykimpel:
I changed the file to reflect the variable_store and included the base_url

thezackm:
> @thezackm regarding the failed tests, is this something for me to fix? I do not seem to have access to Snyk

no need; it's a false positive; you're not pulling in any external dependencies so there should be nothing for Snyk to check.

PR #246, opened 2020-11-21
Added option to hide error_exec output

Dkairu:
Option to hide commands from getting displayed in `error_exec` when set to true output records  COMMAND HIDDEN!

````
integrations:
  - name: nri-flex
    config:
      name: curlFlex
      apis:
        - name: curl
          commands:
            - run: curl -w  -o /dev/null -s https://google.com
              hide_error_exec: true


carlosroman (approved):
LGTM 👍 
Only suggestion is to move the debug statment to before the `if` statment and remove it from L123 as they both look the same. No biggie as just realised we don't test a failure scenario here so something we can come back and clear up another time.

PR #251, opened 2020-12-08
feat: apply custom variables passed through FLEX_META

Kav91:
When FLEX_META environment variable is defined. Eg. `FLEX_META='{"hello":"abc"}'` the json payload is unpacked and attached to global custom attributes. 

FLEX_CMD_PREPEND & FLEX_CMD_APPEND to prepend and append to the command. This will be heavily used with the agents discovery mechanism and provide seamless compatibility to existing Flex configs.


varas (commented):
LGTM, could you please document these features?

Kav91:
@varas updated docs, thanks.

varas (approved):

PR #259, opened 2021-01-04
Thezackm/fix yaml examples

thezackm:
updating windows examples for readability and cleaning up a legacy sample to work with v4 configs


Kav91:
@thezackm switching from `name` to `event_type` will stop it from automatically adding "Sample" at the end of the event type for you. Is that what you also intended to do?

thezackm:
> @thezackm switching from `name` to `event_type` will stop it from automatically adding "Sample" at the end of the event type for you. Is that what you also intended to do?

Yes it is. Nice call-out!

Kav91 (approved):

PR #262, opened 2021-01-12
Added pi-network config

bpschmitt:
Added config to detect USB device and ethernet port connectivity on a Raspberry Pi.

Tested on armv7l GNU/Linux


examples/linux/pi-network.yml
```diff
@@ -0,0 +1,48 @@
...
+# Tested on an armv7l Rapsberry Pi
+---
+integrations:
+  - name: nri-flex
+    config:
+      name: interfaceLinkStatus
+      apis:
+        # Let's get a list of interfaces and save them for the next command
+        - name: interfaceLink
+          event_type: interfaceLookup
```
thezackm, line 12: are both of these attributes needed?

examples/linux/pi-network.yml
```diff
@@ -0,0 +1,48 @@
+# Ensure nri-flex is installed in /var/db/newrelic-infra/newrelic-integrations/bin
+# Place pi-network.yaml in /etc/newrelic-infra/integrations.d and restart the newrelic-infra agent
+# Tested on an armv7l Rapsberry Pi
```
thezackm, line 3: [optional] might be nice to add a sample NRQL for both event types here
e.g.: `SELECT bus, device, id, description FROM usbStatusSample`

It's not very clear to see what the attribute names are from the `interfaceLinkSample` events

examples/linux/pi-network.yml
```diff
@@ -0,0 +1,48 @@
...
+          event_type: interfaceLookup
+          commands:
+            - run: ip link | egrep '^[0-9]' | awk '{gsub(/:/,"",$2); print $2}'
+              set_header: [interface]
+              split: horizontal
+              split_by: ":"
+          ignore_output: true
+        # Use the lookup to loop through the interface list and parse the output
+        - name: interfaceLink
+          event_type: interfaceLinkSample
```
thezackm, line 21: same as above; are both of these required?

thezackm (changes requested):
@bpschmitt - just finished initial review; thanks!

examples/linux/pi-network.yml
bpschmitt: Updated.

examples/linux/pi-network.yml
bpschmitt: Updated.

examples/linux/pi-network.yml
bpschmitt: Updated.

thezackm (approved):
awesome, LGTM!

PR #263, opened 2021-01-14
Kav/command improvements

Kav91:
Adds 3 gated features:

### 1. StdinPipe
	   StdinPipe            bool   `default:"false" help:"use cmd.StdinPipe for commands"`

When enabled stdinPipe will be used for commands, some executables exit immediately causing an exit status 1 with no error. Allowing this functionality allows the command to be run successfully with the expected output.

### 2. FLEX_CMD_WRAP
Automatically wrap a command in quotes (Requires AllowEnvCommands enabled).
This is useful when env prepend or append is being used and the subsequent command needs to be wrapped so it can be executed properly.

### 3. Assert command output before processing
Useful for when example `netstat` is not available on a system but `ss` is, this provides a more gracious way to assert the output and skip if required.

Add an assert block with a nested variable of `match` and/or `not_match`.
Note both options use regex.

Examples:
```yml
#### Command output must contain the string "hi" - this will continue
integrations:
  - name: nri-flex
    config:
      name: SomeIntegration
      apis:
        - event_type: SomeSample
          commands:
            - run: "echo hi:bye"
              split_by: ":"
              assert:
                match: hi #### <--------
```

```yml
#### Command output must contain the string "foo" - this will NOT continue!
integrations:
  - name: nri-flex
    config:
      name: SomeIntegration
      apis:
        - event_type: SomeSample
          commands:
            - run: "echo hi:bye"
              split_by: ":"
              assert:
                match: foo #### <--------
```

```yml
#### Command output must contain the string "hi" and not contain the string "foo - this will continue
integrations:
  - name: nri-flex
    config:
      name: SomeIntegration
      apis:
        - event_type: SomeSample
          commands:
            - run: "echo hi:bye"
              split_by: ":"
              assert:
                match: hi ##### <------------
                not_match: foo #### <--------
```

### DIFF
--- a/docs/apis/commands.md
+++ b/docs/apis/commands.md
@@ -250,9 +250,9 @@ Note both options use regex.
 integrations:
   - name: nri-flex
     config:
-      name: VodafoneSpeedtestIntegration
+      name: SomeIntegration
       apis:
-        - event_type: VodafoneSpeedSample
+        - event_type: SomeSample
           commands:
             - run: "echo hi:bye"
               split_by: ":"
@@ -265,9 +265,9 @@ integrations:
 integrations:
   - name: nri-flex
     config:
-      name: VodafoneSpeedtestIntegration
+      name: SomeIntegration
       apis:
-        - event_type: VodafoneSpeedSample
+        - event_type: SomeSample
           commands:
             - run: "echo hi:bye"
               split_by: ":"
@@ -280,9 +280,9 @@ integrations:
 integrations:
   - name: nri-flex
     config:
-      name: VodafoneSpeedtestIntegration
+      name: SomeIntegration
       apis:
-        - event_type: VodafoneSpeedSample
+        - event_type: SomeSample
           commands:
             - run: "echo hi:bye"
               split_by: ":"
--- a/docs/apis/commands.md
+++ b/docs/apis/commands.md
@@ -2,11 +2,11 @@
 
 The `commands` API allows you to retrieve information from any application or shell command. It can accept multiple commands executed in sequence, raw text, and JSON data, and is platform-agnostic. Keep in mind that platform specific applications differ between different operating systems, and some may not be available (for example, Kubernetes).
 
-* [Basic usage](#Basicusage)
-* [Configuration properties](#Configurationproperties)
-* [Advanced usage](#Advancedusage)
+- [Basic usage](#Basicusage)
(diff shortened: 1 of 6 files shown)


docs/apis/commands.md
```diff
@@ -236,3 +239,54 @@ apis:
 ```
 
 Note that `line_end` is exclusive, meaning that you have to add `1` (`0` indexed) to the actual line you want to stop being processed.
+
+### Assert output exists before processing
+
+Add an assert block with a nested variable of `match` and/or `not_match`.
+Note both options use regex.
```
Noly, line 246: Could you add examples that use regex?

Kav91: @Noly the example is regex already, the output is `hi:bye` the match value is `hi` so because it is regex it matches and continues. If it was not regex that assertion and example would not continue as documented.

internal/inputs/commands.go
```diff
@@ -105,8 +106,30 @@ func commandRun(dataStore *[]interface{}, yml *load.Config, command load.Command
...
+				"exec":       command.Run,
+				"err":        err,
+				"suggestion": "StdinPipe failed",
+			}).Debug("command: failed")
+		}
+	}
+
 	output, err := cmd.CombinedOutput()
 
+	// check assertion, if the result is false return immediately
```
Noly, line 124: Rathen than comment _what_ is doing could you comment _why_ is doing what is doing.

Kav91: @Noly have updated both the comments on that line, and put further detail above the function itself.

docs/apis/commands.md
Noly: I see, I'm curious on what would happen if customer set up smt like: 
 `match: !hi `

It would continue with all outputs that doesn't match with hi? 🤔

Kav91: `!hi` would not match, and the assertion would fail.

Also the way this regex is defined is implemented the same way as the other existing features in Flex that also use regex. 
So for users the experience will ultimately be consistent and the same for them in other places of Flex as well.

Noly: 👍 Make sense. Thanks for the explanation!

Noly:
LGTM

Thanks for implementing this new features and taking time to clear my doubts. 
I'll approve it!

Noly (approved):

PR #272, opened 2021-02-01
Fix: parse_html - strip special chars from html <Table> element

haihongren:
Strip following special chars from html `<Table>` element, to prevent illegal JSON chars. 

   - ('\t', '\n', '\v', '\f', '\r', ' ', U+0085 (NEL), U+00A0 (NBSP)
   - category P unicode (Punctuation)


Kav91 (approved):
nice fix!

PR #276, opened 2021-02-16
feat: allow custom storer name and storer ttl 

haihongren:
### Custom storer name and ttl using environment variables 
- STORER_NAME 
- STORER_TTL 
 
By default, the Infra integration SDK uses the integration name as the storer name. For cases where multiple instances of Flex running concurrently, writing to the same storer can result in race condition. Setting a unique storer name using `STORER_NAME` for each instance overcomes this issue. 

The integration SDK uses 1m as the storer TTL by default.  For integrations with longer than 1m polling interval, the data stored with storer will be void by the next interval. This causes issue calculating DELTA and RATE metrics.  Setting longer TTL( than the polling interval) using `STORER_TTL` solves this issue.

> Sample integration with `STORER_NAME` and `STORER_TTL` usage
```yaml
integrations:
  - name: nri-flex
    interval: 24h
    env:
      CONFIG_FILE: /etc/newrelic-infra/integrations.d/universal-discovery-sub-configs/iostats-linux-vmware.yml   
      STORER_NAME: ${discovery.vmName}      
      STORER_TTL: 5m5s
```


Kav91 (approved):
Awesome!

PR #286, opened 2021-02-25
fix: context errors not captured

Kav91:
When an error occurs during command execution, the code returns early and does not capture the context error. 
Things like command timeouts - context deadline exceeded are not captured.

cc/ @haihongren


carlosroman (approved):
LGTM 👍🏼

Noly (commented):
Thanks @Kav91 for taking care of this issue.
Could it be possible to cover it with unit-test?

PR #291, opened 2021-03-04
Leons mysql flex

lchapman4:
Added an entry to pull the server variables into NR.  This allows a customer to compare current connections against the configured max connections which is a key metric to monitor.


lchapman4:
ugh, wrong file here in this PR.  Ignore this one.  I'll resubmit with the correct file.

PR #292, opened 2021-03-05
Leons mysql flex

lchapman4:
Updated the sample file to pull mysql variables and gather max connections with other config values.


examples/mysql-database-example.yml
```diff
@@ -5,17 +5,57 @@ integrations:
...
           db_async: true # process queries async
           db_queries:
-            - name: mysqlSessionSample
-              run: Select * from sys.dm_exec_sessions
+            - name: MySQLShowVars
+              run: SHOW VARIABLES;
               custom_attributes: # can apply additional at a nested level
-                nestedAttr: nestedVal
+                nestedAttr: max-conns
+            - name: MySQLCustomQuery
```
thezackm, line 23: should this be named something more relative to the query similar to the examples on lines 19 and 27?

thezackm (changes requested):

examples/mysql-database-example.yml
lchapman4: Good point. I'll update the names to be more consistent.

thezackm (approved):

PR #299, opened 2021-05-05
Adding MarkLogic DB Flex Examples

shredtechular:
Adding 2 example yml configurations for monitoring [MarkLogic](https://www.marklogic.com/) using various API endpoints they expose in their management API.

- One example is for MarkLogic instances that have Basic Authentication configured on it's API.
- The other example is for MarkLogic instances that have DigestAuthentication configured on it's API.


Kav91 (approved):
LGTM nice work 👍🏾

PR #306, opened 2021-06-09
fix: if multiple jq results exist return all

Kav91:
When multiple JQ results are returned, only the first is returned causing the loop to exit immediately.
PR stores each result in a slice and eventually returns when the loop is complete.


rubenruizdegauna (approved):
LGTM

rubenruizdegauna:
Hi @Kav91 , I made a test to check the change. Would you add it? (one or different one if you prefer)
```golang
func TestRunJqMultipleResults(t *testing.T) {

	input := []interface{}{map[string]interface{}{
		"foo": map[string]interface{}{
			"data": map[string]interface{}{
				"abc": 1,
				"def": 2,
			},
			"more_data": map[string]interface{}{
				"ghi": 3,
				"jkl": 4,
			},
		},
	}}

	api := load.API{
		Jq: ".foo.data,.[0].foo.more_data",
	}

	var dataSets = runJq(input, api)
	var expectedResult = []interface{}{
		map[string]interface{}{"abc": 1, "def": 2},
		map[string]interface{}{"ghi": 3, "jkl": 4},
	}

	assert.Equal(t, expectedResult, dataSets)
}

```

thanks!

Kav91:
@rubenruizdegauna good catch, thanks for the test! I added your suggested one.

PR #320, opened 2021-09-09
Update ebpf-tcplife.yml

nethippo:
split has 11 "(\S+)"  but headers have 12. I have added one more ",(\S+)" to make it full.


rubenruizdegauna (approved):
LGTM

PR #321, opened 2021-09-14
example config to ingest metrics for Active Directory

GaPeppy:
example config based on windows-ps-diskperf for Active Directory; tested on Window Server 2019


brushknight (approved):
LGTM

PR #334, opened 2022-02-06
Add example on how to use Pixie CLI and Flex to push k8s metrics into NRDB 

nvhoanganh:


examples/pixieExample/README.md
```diff
@@ -0,0 +1,183 @@
+# Collect custom Pixie metrics from your k8s cluster and push to New Relic for Alerting and Dashboarding
+
+The following documentation explains how to configure the New Relic FLEX integration to collect custom Pixie metrics from your k8s cluster.
```
Kav91, line 3: Rename FLEX to Flex to match all product references.

examples/pixieExample/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+-   Golden metrics (throughput, latency, error rate) for HTTP-based services
+-   HTTP transaction data
+-   Database transaction data (for MySQL & PostgreSQL)
+-   Distributed tracing
+-   JVM metrics
+
+With New Relic Flex, you now can report and alert on all other available Pixie metrics by:
+
+-   write custom [Pixie script](https://docs.pixielabs.ai/tutorials/pxl-scripts/write-pxl-scripts/)
```
Kav91, line 17: We should be using sentence case for these dot points. So just use a capital at the start of each bullet point, can see there are a few of these below.

Kav91 (changes requested):
This is a great solution and guide, nice work!
I think we should also have a reference to this on the main README.md, I'm thinking somewhere under the documentation section here -> https://github.com/newrelic/nri-flex#documentation .

examples/pixieExample/README.md
nvhoanganh: done

nvhoanganh:
> This is a great solution and guide, nice work!
> I think we should also have a reference to this on the main README.md, I'm thinking somewhere under the documentation section here -> https://github.com/newrelic/nri-flex#documentation .

I will create another PR to modify the main README file once this is merged

examples/pixieExample/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+RUN ./install-pixie-no-auth-docker.sh
+```
+
+-   Push the docker image to your docker registry
+
+## Step 2: Configure required Pixie secrets
+
+```bash
+# create file containging New Relic ingest API key
```
Kav91, line 42: typo `containging`

Kav91 (approved):

PR #339, opened 2022-02-14
fix: dial connection close() handling when context timeout

haihongren:
Fixed: 
- Connection is not closed when context timeout happens

### DIFF
--- a/internal/inputs/netdial.go
+++ b/internal/inputs/netdial.go
@@ -57,8 +57,13 @@ func NetDialWithTimeout(dataStore *[]interface{}, command load.Command, dataSamp
 				reader := bufio.NewReader(dialConn)
 				tp := textproto.NewReader(reader)
 				for {
-					line, _ := tp.ReadLine()
-					data += line + "\n"
+					select {
+					case <-ctx.Done():
+						return
+					default:
+						line, _ := tp.ReadLine()
+						data += line + "\n"
+					}
 				}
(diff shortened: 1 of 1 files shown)


internal/inputs/netdial.go
```diff
@@ -60,7 +63,7 @@ func NetDialWithTimeout(dataStore *[]interface{}, command load.Command, dataSamp
 			}
 		}
 		c <- struct{}{}
```
cristianciutea, line 65: In case the context is stoped, I think that this go routine will "leak".
I was thinking that maybe instead of those changes, making the chan `c` a buffered chan of length 1, will fix the issue with dialConn not being closed. WDYT?

cristianciutea (changes requested):

internal/inputs/netdial.go
haihongren: @cristianciutea thanks for pointing out the go routine "leak".  I tried buffered chan of length 1. It did not fix the dialConnect not being closed issue. 

I made a new update to avoid the go routine "leak" issue. Please have a look.  ta

cristianciutea: I've just saw how we call ReadLine(), you're right using a buffered channel will not fix the issue.
The latest commit will fix the go routine leak. Thank you!

cristianciutea (approved):

PR #342, opened 2022-03-02
ci: run tests on pull requests

rogercoll:


.github/workflows/push_pr.yml
```diff
@@ -11,6 +11,7 @@ on:
       - README.md
       - 'docs/**'
       - 'examples/**'
+  pull_request:
```
rubenruizdegauna, line 14: we need to discuss this a bit further as if the secrets we are using in the repo are not present in the fork it will not work.

rogercoll: You are totally right! We might need to separate testing workflows with the ones that need secrets (login) like snyk. 

Thanks for pointing that out

rogercoll:
Moved to draft to decouple workflows that require secrets, thus testing workflows could be run from external forks without any problem.

https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

`Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.`

PR #348, opened 2022-03-28
URL checking example using curl

GaPeppy:
This example showcases a http-layer test with the intent of checking web endpoints on localhost.  The associated urlcheck.json has public web endpoints that should be replaced with localhost endpoints.


rubenruizdegauna (approved):
LGTM

rubenruizdegauna:
Thanks for your contribution @GaPeppy

PR #361, opened 2022-05-15
added an example for getting azure vm core count usage for a location

Nimzr:
Added a Azure VM Folder and an example on how to get usage details.  
Eventually I may add an example of doing this with an azure function but it was easier to get started as a flex.


rubenruizdegauna (approved):
LGTM

PR #382, opened 2022-09-30
okla speedtest metrics using the JSON output

adatole:
This integrates internet speed metrics coming out of the Okla speedtest utility (which needs to be running locally on the target system) into New Relic, which can then be added to dashboards.

### DIFF
--- a/examples/linux/okla-speedtest.yml
+++ b/examples/linux/okla-speedtest.yml
@@ -3,7 +3,6 @@ integrations:
     timeout: 5m
     interval: 1h
     config:
-      
       name: linuxspeedtest 
       apis:
         - name: speedtest
--- a/examples/linux/okla-speedtest.yml
+++ b/examples/linux/okla-speedtest.yml
@@ -1,13 +1,14 @@
 integrations:
   - name: nri-flex
     timeout: 5m
+    interval: 1h
     config:
-      interval: 120s
+      
       name: linuxspeedtest 
       apis:
         - name: speedtest
           commands:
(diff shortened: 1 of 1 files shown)


rubenruizdegauna:
Hi @adatole , thanks for your contribution! 

I have a few comments about this:
* I tried the command on an ubuntu-20.04 and `--accept-license` didn't work for me. It gets stuck waiting for approval. I didn't see it in the command options. Is it still supposed to work?
* I tried w/o sudo and it works, so I don't think it is required.
* Speedtest consumes some bandwidth and if it's executed ever 20'' it will take too much bandwith IMO. I would add a bigger interval (i.e. every hour)


Also the commits should be signed 🙂  
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

adatole:
THANK YOU so much for the guidance.

- I tested the --accept-license command on a clean system and verified it works. 
- My tests on a clean system indicate sudo is needed and fails with "exit status 134" without it.
- upped the interval to 1h (and put it in the correct location)
- resubmitted with a signed version (I think).

rubenruizdegauna (approved):
LGTM

rubenruizdegauna:
> THANK YOU so much for the guidance.
> 
> * I tested the --accept-license command on a clean system and verified it works.
> * My tests on a clean system indicate sudo is needed and fails with "exit status 134" without it.
> * upped the interval to 1h (and put it in the correct location)
> * resubmitted with a signed version (I think).

Great! Thank you for the contribution :)

PR #383, opened 2022-10-17
NR-57870 fix: check assertion with and without match

rogercoll:


test/testbed/scenarios/fixtures/command_api.go
```diff
@@ -61,15 +61,15 @@ integrations:
       apis:
         - event_type: LinuxFileList
           commands:
-            - run: ls -l /etc/apt/
+            - run: ls -l --time-style="+%Y" /etc/apt/
```
rogercoll, line 64: E2E tests where failing because the displayed data, ls date format varies depending on the local time: https://unix.stackexchange.com/questions/237531/why-does-ls-all-show-time-for-some-files-but-only-year-for-others

rubenruizdegauna (approved):
nice!

PR #384, opened 2022-10-19
feat: add value_to_upper & value_to_lower + tests

Kav91:
Adds the two below features with accompanying tests.
This is for an active customer requirement.

---

### value_to_lower

Finds keys and convert all their values to lowercase

**Example**

```json
{
  "id": "eca0338f4ea31566",
  "name": "NODE"
}
```

You could use `value_to_lower` to transform the value to lower case.

```yaml
name: example
apis:
  - name: someService
    value_to_lower:
      - name
```

Which would return the following:

```json
"metrics": [{
  "event_type": "someServiceSample",
  "id": "eca0338f4ea31566",
  "name": "node",
}]
```

### value_to_upper

Finds keys and convert all their values to uppercase

**Example**

```json
{
  "id": "eca0338f4ea31566",
  "name": "node"
}
```

You could use `value_to_upper` to transform the value to lower case.

```yaml
name: example
apis:
  - name: someService
    value_to_lower:
      - name
```

Which would return the following:

```json
"metrics": [{
  "event_type": "someServiceSample",
  "id": "eca0338f4ea31566",
  "name": "NODE",
}]
```


rogercoll:
Hi @Kav91  ,

Could you add some end-to-end tests in https://github.com/newrelic/nri-flex/tree/master/test/testbed that cover this use case?

Thanks for the contribution!

Kav91:
@rogercoll from the time I created Flex we had always used the standard go testing pattern. I can definitely see the value of using this new testbed however this functionality and contributing guidelines are not documented anywhere, I'm not even sure if we should still do the standard go tests over using this testbed now either. 

My point here is that anything that deviates from standard development and contribution practice like using this testbed over or with standard tests should be documented so contributors are aware. 

Can you update the contributing guidelines to document this, and can you also clarify if standard tests are still required or just going forward with updating the testbed only, or both.

rogercoll:
@Kav91 Good point! I just created a task for the team to do so. Thanks for pointing that out.

Kav91:
@rogercoll E2E tests added they are passing, so should be good to merge unless you spot anything else. 
The error here is for some other failing test completely unrelated to my PR.

rubenruizdegauna (approved):
LGTM

PR #397, opened 2022-11-28
Update linux-systemctl-cmd-example.yml

travisroberts-sf:
by adding the --all flag you will get the status of services when they actually go down.  Otherwise, the service falls off the list entirely.


rubenruizdegauna (approved):
LGTM

PR #398, opened 2022-11-29
feat: support aws ssm parameter secrets

hans-m-song:
Implementation of [NEWRELIC-5592](https://issues.newrelic.com/browse/NEWRELIC-5592)

A POC for AWS SSM support, following how AWS KMS was implemented.

### enhancement

- Support fetching AWS SSM parameters for integration variables

### use case

AWS SSM is used much a much more cost-effective and flexible solution to store parameters as opposed to AWS KMS. This would help those who would like to move integrations from AWS into nri-flex but want to keep secrets/parameters in AWS. Those running the infrastructure agent in an EKS cluster can configure service accounts with access to parameters, simplifying access.

### example

```yaml
variables:
  tag_name:
    aws-ssm:
      region: ap-southeast-2
      type: plain
      data: /path/in/ssm

integrations:
  - name: nri-flex
     config:
       name: AwsCostExplorerIntegration
       apis:
         - event_type: AwsCostExplorerSample
            commands:
              - run: |
                   aws ce get-cost-and-usage \
                   --group-by Type=TAG,Key="${tag_name}"
                   ...
```

### todo

- [ ] tests
- [ ] documentation
- [ ] how to consume parameter path?
- [ ] how to specify whether or not to decrypt the parameter value

### DIFF
--- a/internal/config/secrets.go
+++ b/internal/config/secrets.go
@@ -20,6 +20,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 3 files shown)


internal/config/secrets.go
```diff
@@ -57,6 +58,8 @@ func loadSecrets(config *load.Config) error {
 		results := map[string]interface{}{}
 
 		switch secret.Kind {
+		case "aws-ssm":
+			secretResult = awsssmFetch(name, tempSecret)
 		case "aws-kms":
 			if secret.Region == "" {
```
hans-m-song, line 64: is this check actually too stringent? the region is only used if config files are not provided

hans-m-song:
FWIW, the work related to this feature has been dropped so feel free to close.

PR #399, opened 2022-11-29
chore: example of including aws cli

hans-m-song:
An RFC for extending the base image with additional binaries (in this case AWS CLI).

Using ubuntu as the base image was important as AWS CLI requires glibc and is incompatible with musl

Would be good to get some thoughts/feedback.


muruganishekar:
@hans-m-song please sign CLA and and raise a PR

Issue #402, opened 2022-12-22
"lookups" function fails when looking up more than one value in same command


abhisheklad:
## Description
[lookups](https://github.com/newrelic/nri-flex/blob/master/docs/basics/functions.md#lookups) function is failing when used to lookup more than one values in same call/command.

## Expected Behavior
looups function replaces more than one value as required. Behaviour will be exactly same as looking up a single value.

### Verbose Logs
```
 $  ./bin/darwin/nri-flex -config_file examples_local/lookuptest.yml -pretty -verbose
DEBU[0000] Function.isAvailable: enter
DEBU[0000] Function.isAvailable: exit status: false
INFO[0000] com.newrelic.nri-flex                         GOARCH=arm64 GOOS=darwin version=Unknown-SNAPSHOT
DEBU[0000] config: git sync configuration not set
WARN[0000] config: testing agent config, agent features will not be available
DEBU[0000] config: running async                         name=LookUps
DEBU[0000] config: processing apis                       apis=2 name=LookUps
DEBU[0000] fetch: collect data                           name=LookUps
DEBU[0000] processor-data: running data handler          name=LookUps
DEBU[0000] processor-data: running data handler          name=LookUps
DEBU[0000] fetch: collect data                           name=LookUps
ERRO[0000] fetch: failed to process file                 error="file input: failed to read file: open examples_local/${lookup.readSample:brand}_${lookup.readSample:car}.json: no such file or directory" file="examples_local/${lookup.readSample:brand}_${lookup.readSample:car}.json" name=LookUps
DEBU[0000] processor-data: running data handler          name=LookUps
DEBU[0000] config: finished variable processing apis     apis=2 name=LookUps
INFO[0000] flex: completed processing configs            configs=1
{
	"name": "com.newrelic.nri-flex",
	"protocol_version": "3",
	"integration_version": "Unknown-SNAPSHOT",
	"data": [
		{
			"metrics": [
				{
					"brand": "honda",
					"car": "civic",
					"event_type": "readSample",
					"integration_name": "com.newrelic.nri-flex",
					"integration_version": "Unknown-SNAPSHOT"
				},
				{
					"brand": "toyota",
					"car": "supra",
					"event_type": "readSample",
					"integration_name": "com.newrelic.nri-flex",
					"integration_version": "Unknown-SNAPSHOT"
				},
				{
					"brand": "mistsubishi",
					"car": "lancer",
					"event_type": "readSample",
[...]

PR #403, opened 2022-12-30
|NEWRELIC-6016 fix: allow multiple values in lookup

rubenruizdegauna:
Allow accessing multiple values from lookup output

Fixes https://github.com/newrelic/nri-flex/issues/402


brushknight (approved):
LGTM

PR #406, opened 2023-01-22
fix: doc update, Anchor link fix for sections

zmrfzn:


Kav91:
good catch, thanks 👍🏾

PR #413, opened 2023-02-17
updated and also new pihole integration

LeonAdato:
I've locked down the _simple version, and added a _detailed version of the flex integration for monitoring the pihole.


examples/linux/okla-speedtest.yml
```diff
@@ -0,0 +1,17 @@
+integrations:
```
rogercoll, line 1: This file already exists in a different path: https://github.com/newrelic/nri-flex/blob/master/examples/okla-speedtest.yml

Could you move it to the desired path? `examples/linux/oka-speedtest.yml`?

rogercoll (changes requested):

examples/linux/okla-speedtest.yml
LeonAdato: Done.

PR #415, opened 2023-02-21
changed the days checked from 300 to 30 and added a basic overview

Nimzr:
Changed the number of days that the script checks from 300 to 30 and added a basic readme in preparation of requesting it to be added to Instant Observability.


rogercoll (approved):
LGTM

PR #417, opened 2023-03-13
chore(deps): bump net package version

rogercoll:


build/Dockerfile
```diff
@@ -9,4 +9,6 @@ COPY tools /app/tools
 
 COPY go.mod go.sum Makefile /app/
 
+ENV GOFLAGS="-buildvcs=false"
```
rubenruizdegauna, line 12: is this necessary ☝️ ?  (just ensuring is not a leftover from local)

rogercoll: Yes, is the same issue that happened with the agent a few weeks ago: https://github.com/newrelic/nri-flex/actions/runs/4404183693/jobs/7713478560#step:4:184

rogercoll: Goreleaser fails with new Go container image due VCS and project's name. (we need to bump GoReleaser version too)

rubenruizdegauna (approved):
LGTM 🔥

PR #421, opened 2023-03-24
NR-100933 chore: upgrade to Go 1.20

DavSanchez:
- NR-100933 chore: upgrade to Go 1.20
- chore: upgrade GHA workflows


rogercoll (approved):
LGTM

PR #422, opened 2023-04-06
feat:add single flex integration metric example

ismahernandez:
Adding an example file to show how a single metric coming out of a file can be defined using the Flex integration.
In the example, it reads the temperature of the cpu. (It depends on the system where this temp is to be found at)


Kav91 (approved):
LGTM

PR #429, opened 2023-05-24
feat: add sqlite db driver

Kav91:
Adds SQLite DB driver.

Note this functionality does not effect any core supported features.


Kav91:
Given this works identical to our existing DB driver feature and is simply a drop-in, as well as still being classed as experimental. Will merge this to master as there are no direct support related impacts or potential impact to any existing supported features.

PR #441, opened 2023-07-25
ci/use goreleaser github action

DavSanchez:
- ci: use goreleaser action instead of make target


DavSanchez:
This was relatively straight forward, but the make target makes easy to run this locally, and it will be cumbersome to maintain both options (the one not linked to the CI would probably lag behind and cause trouble when used). We prefer to leave the make target and run it as is from the CI.

Closing.

PR #447, opened 2023-09-20
Linux: Informix onstat utility usage

andrew-lozoya:
Demonstrates the usage of the IBM® Informix onstat utility


andrew-lozoya:
image005

Conducted a valid test in a lab and with the customer, and found that onstat was somewhat challenging because flex was not accepting raw integers as value captures. This illustrates how to utilize regex to extract integers from wrapped string output.

rubenruizdegauna (approved):
LGTM

PR #462, opened 2024-02-14
Bump SDK to 3.8.2 and allow TempDir

alvarocabanas:
- Bump SDK to 3.8.2 and allow TempDir modification for caches
- Adapt newMetricSet add namespace Attributes

### DIFF
--- a/internal/outputs/infra.go
+++ b/internal/outputs/infra.go
@@ -76,5 +76,8 @@ func createStorer() (persist.Storer, error) {
 	if os.Getenv("VERBOSE") == "true" || os.Getenv("VERBOSE") == "1" {
 		fmt.Fprintf(os.Stderr, "custom storer name: %s and TTL: %s\n", storerName, ttl)
 	}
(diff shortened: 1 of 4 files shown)


internal/outputs/infra.go
```diff
@@ -76,5 +76,8 @@ func createStorer() (persist.Storer, error) {
 	if os.Getenv("VERBOSE") == "true" || os.Getenv("VERBOSE") == "1" {
 		fmt.Fprintf(os.Stderr, "custom storer name: %s and TTL: %s\n", storerName, ttl)
 	}
-	return persist.NewFileStore(persist.DefaultPath(storerName), load.Logrus, ttl)
+
+	tempDir := os.Getenv("TEMP_DIR")
```
rogercoll, line 80: If TEMP_DIR is undefined, will it use the old default storer path?

rogercoll (approved):
LGTM

internal/outputs/infra.go
alvarocabanas: Yes, if it's undefined the call to  TmpPath will call to DefaultPath.

PR #463, opened 2024-02-26
Added support for SAP Sybase Database.

miransar:


Kav91 (approved):
LGTM
Given this works identical to our existing DB driver feature and is simply a drop-in, as well as still being classed as experimental. Will merge this to master as there are no direct support related impacts or potential impact to any existing supported features.

PR #496, opened 2024-05-13
Updated the MSSQL driver to use an Azure tokens

Dkairu:
The current MSSQL driver uses username and password to authenticate but not enough options are offered to authenticate against MS Entra ID formerly Active directory…This PR gives additional options like using tokens that are dynamically generated from Entra ID/ Active Directory instead of user password to provide flexibility of using the integration.


Kav91:
Is this backwards compatible with the existing implementation? I can't see anything in the main readme confirming this.

Dkairu:
@Kav91 Yes as a matter of fact it uses the modules from the original driver https://github.com/denisenkom/go-mssqldb/blob/master/azuread/driver.go. I did test the user name and password and it works.

PR #514, opened 2024-07-31
chore: bump deps

gsanchezgavier:
Bump dependencies 

github.com/aws/aws-sdk-go v1.55.4 -> v1.55.5
github.com/docker/docker v25.0.6+incompatible -> v26.1.5+incompatible
github.com/itchyny/gojq v0.12.12 -> v0.12.16
github.com/parnurzeal/gorequest v0.2.15 -> v0.3.0
github.com/prometheus/client_model v0.4.0 -> v0.6.1
github.com/prometheus/common v0.43.0 -> v0.55.0
go.uber.org/ratelimit v0.2.0 -> v0.3.1
golang.org/x/crypto v0.23.0 -> v0.25.0
golang.org/x/net v0.25.0 -> v0.27.0


internal/discovery/discovery.go
```diff
@@ -40,7 +41,7 @@ func Run(configs *[]load.Config) {
 			}).Error("discovery: unable to set docker client")
 		} else {
 			ctx := context.Background()
-			containerList, err := cli.ContainerList(ctx, types.ContainerListOptions{})
+			containerList, err := cli.ContainerList(ctx, container.ListOptions{})
```
alvarocabanas, line 44: Do we have a test for this?

alvarocabanas (commented):
I left a comment.

internal/discovery/discovery.go
gsanchezgavier: I don't think so, there is a non implemented unit test, and I don't find any other tests, pretty sad TBH. 
Even though that , this is just change implies no more risk that any other version bump that is being done with automerge, since the type was moved from one version to another.

```
// ContainerListOptions holds parameters to list containers with.
//
// Deprecated: use [container.ListOptions].
type ContainerListOptions = container.ListOptions
```

alvarocabanas (approved):

PR #518, opened 2024-08-30
Add: more examples

pnvnd:


examples/ibmmq.yml
```diff
@@ -0,0 +1,23 @@
+integrations:
+  - name: nri-flex
+    interval: 30s
+    config:
+      name: ibmmq
```
gsanchezgavier, line 5: is this telemetry something that could be also scraped using the [ibmmq integration ](https://docs.newrelic.com/install/ibm-mq/)? If so , there is any difference on the data collected ?

gsanchezgavier (approved):
Thanks! LGTM just left a question . BTW CLA needs to be signed before merge

muruganishekar:
@pnvnd please raise a new PR with verified commits, make sure to sign the CLA

PR #530, opened 2024-11-06
NR-291019 add fips compliant package builds

rajrohanyadav:
## What?
Adds compilation and packaging of FIPS compliant packages using `boringcrypto`
## Why?
Many customers can only use FIPS compliant packages in their environments
## How?
* use a new Dockerfile to build FIPS and non-FIPS packages
* update goreleaser config to build and package FIPS compliant packages in addition to the normal packages
* Update release workflow to use the same format as infra-agent for coverage.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,6 +1,44 @@
-ARG GO_VERSION=1.22
+# Use Ubuntu 16.04 as the base image
+FROM ubuntu:16.04
 
-FROM golang:$GO_VERSION
+# Define Go version
+ARG GO_VERSION=1.22
+ARG ARCH='amd64'
+
+# Install dependencies
+RUN apt-get update && apt-get install -y \
+    curl \
+    wget \
+    expect \
+    git \
+    tar \
+    gcc \
+    jq \
+    g++ \
+    gnupg2 \
+    gnupg-agent \
+    debsigs \
+    rpm \
+    build-essential \
+    software-properties-common \
+    python-software-properties \
+    gcc-arm-linux-gnueabi \
+    dpkg-sig \
+    gcc-aarch64-linux-gnu
+
+# Install Go
+# It tries to get the latest patch version of Go that matches the version specified in the Makefile
+# It can be simplified once we have Renovatebot support for this repo
+RUN FULL_GO_VERSION=$(curl -s "https://go.dev/dl/?mode=json" | \
+    jq -r --arg VERSION "${GO_VERSION}" \
+    '[.[] | select(.version | test("^go\($VERSION)\\."))] | .[0].version') && \
+    curl -sSL https://golang.org/dl/${FULL_GO_VERSION}.linux-${ARCH}.tar.gz -o ${FULL_GO_VERSION}.linux-${ARCH}.tar.gz && \
+    tar -C /usr/local -xzf ${FULL_GO_VERSION}.linux-${ARCH}.tar.gz && \
+    rm ${FULL_GO_VERSION}.linux-${ARCH}.tar.gz
+
+# Set Go environment variables
+ENV PATH="/usr/local/go/bin:${PATH}"
+ENV GOPATH="/go"
 
 WORKDIR /app
 
@@ -9,7 +47,10 @@ COPY tools /app/tools
 
 COPY go.mod go.sum Makefile /app/
 
+# Optional: Set Go environment flags
 ENV GOFLAGS="-buildvcs=false"
+
+# Optional: Configure git
 RUN git config --global --add safe.directory /go/src/github.com/newrelic/nri-flex
 
 RUN make deps
--- a/build/compile.mk
+++ b/build/compile.mk
@@ -19,7 +19,7 @@ compile-only: deps-only
 	@mkdir -p $(BUILD_DIR)/$(GOOS)
 	@for b in $(BINS); do \
 		echo "=== $(PROJECT_NAME) === [ compile          ]:     $(BUILD_DIR)$(GOOS)/$$b"; \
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ SRCDIR           ?= .
 BUILD_DIR        ?= $(CURDIR)/bin
 COVERAGE_FILE    ?= coverage.out
 
-GO_VERSION       ?= 1.22
+GO_VERSION       ?= 1.22.8
```
gsanchezgavier, line 9: until renovate an automations are in place i would just pinned the major so we get the patch on every release.

build/Dockerfile
```diff
@@ -1,6 +1,38 @@
-ARG GO_VERSION=1.22
+# Use Ubuntu 16.04 as the base image
+FROM ubuntu:16.04
 
-FROM golang:$GO_VERSION
+# Define Go version
+ARG GO_VERSION=1.22.8
```
gsanchezgavier, line 5: would be possible to avoid having multiple declarations of the go version, it could be prone to error whenever we need to bump it.

build/compile.mk
```diff
@@ -40,10 +40,21 @@ compile-linux: deps-only
...
 
+.PHONY: compile-linux-fips
+compile-linux-fips: deps-only
+	@echo "=== $(PROJECT_NAME) === [ compile-linux-fips    ]: building commands:"
+	@mkdir -p $(BUILD_DIR)/linux-fips
+	@for b in $(BINS); do \
+		OUTPUT_FILE="$(BUILD_DIR)linux-fips/$$b" ; \
+		echo "=== $(PROJECT_NAME) === [ compile-linux-fips    ]:     $$OUTPUT_FILE"; \
+		BUILD_FILES="$(SRCDIR)/cmd/..." ; \
+		GOOS=linux GOFIPS=1 CGO_ENABLED=1 CC=aarch64-linux-gnu-gcc GOEXPERIMENT=boringcrypto $(GO_CMD) build -tags fips -ldflags="$(LDFLAGS)" -o $$OUTPUT_FILE $$BUILD_FILES ; \
```
gsanchezgavier, line 55: I have not context on this but i assume that is correct, perhaps @alvarocabanas  could double check

cmd/nri-flex/nri-flex-fips.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2019 New Relic Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build fips
+// +build fips
+
+package main
```
gsanchezgavier, line 7: same doubt i had here https://github.com/newrelic/infrastructure-agent/pull/1942#discussion_r1827743835

gsanchezgavier (commented):
thanks just left a couple of comments, i see that there is no workflow to run this neither push the artifacts. i guess that it would be following prs ?

Makefile
rajrohanyadav: Needed the patch to download and install in the Dockerfile, but changed it to fetch the latest go patch version based on `major.minor`. Can be simplified once we have renovate. Thanks for pointing it out.

build/Dockerfile
rajrohanyadav: This version is only the default version to be used in case we don't get a version from Makefile. But https://github.com/newrelic/nri-flex/blob/5c2e540fc3223c786e9200390b92e0b2b14b9965/build/ci.mk#L7 specifies a build arg which comes from the Makefile.

rajrohanyadav:
> thanks just left a couple of comments, i see that there is no workflow to run this neither push the artifacts. i guess that it would be following prs ?

Thanks for the comments. I believe goreleaser takes care of pushing the artifacts to the release from the runner (as happening until now for this repo). I don't know if we need to publish these to S3. So, there would be no changes to the way packages are published, just two additional packages with be created for each release.

cmd/nri-flex/nri-flex-fips.go
rajrohanyadav: > same doubt i had here [newrelic/infrastructure-agent#1942 (comment)](https://github.com/newrelic/infrastructure-agent/pull/1942#discussion_r1827743835)

Addressed. Thanks for pointing it out.

gsanchezgavier: have you tried by just adding an extra file only for fips? no main no nothing? I'm not sure if would work here but a minimal example on my environment works

alvarocabanas: Yes, I was going to  comment the same, you can do as Guillermo says, leaving nri-flex as it is and just adding a new nri-flex-fips file with:

```

//go:build fips
// +build fips

package main

import (
	_ "crypto/tls/fipsonly"
)

```

build/compile.mk
```diff
@@ -40,10 +40,21 @@ compile-linux: deps-only
 	@for b in $(BINS); do \
 		OUTPUT_FILE="$(BUILD_DIR)linux/$$b" ; \
```
alvarocabanas, line 41: We have this legacy error in all the compile targets, it should be OUTPUT_FILE="$(BUILD_DIR)/linux/$$b" and the same for every os, since there is @mkdir -p $(BUILD_DIR)/linux just over creating the right structure but the binary gets-created in the folder binlinux/

alvarocabanas (commented):
I left some comments

build/compile.mk
rajrohanyadav: I fixed it locally but then forgot to push :D. Thanks for pointing out. Fixed it.

alvarocabanas (approved):
LGTM, GOOD JOB!

PR #537, opened 2024-12-17
fix(fips): update fips archive name format

rajrohanyadav:
To unify the naming format for archives and packages, updating `fips` archive names.
Also updated `goreleaser` as part of the same PR.


alvarocabanas (approved):
LGTM

PR #555, opened 2025-03-06
ci: fix builds because of missing go1.22 package

ujwalsah:
After the release the go version 1.24, the go version 1.22 is no longer supported. Hence, hardcoded the go download URL for now to point to go version 1.22.


rajrohanyadav (approved):
LGTM

rajrohanyadav:
Thanks for fixing this, as you pointed out we'll have to find a path forward for the deprecated go version. It is a concern as we won't get any security updates from go team :(

sairaj18 (commented):
Thanks for working on this 😃.

I'm assuming [this](https://github.com/newrelic/nri-flex/actions/runs/13693597727/job/38291103899?pr=555) PR check will be fixed by this [PR](https://github.com/newrelic/nri-flex/pull/546/files)

sairaj18 (approved):

PR #561, opened 2025-03-18
fix: replace proper env var rather than greedily replacing substr

rajrohanyadav:
Given two env vars: `ABC` and `ABC_DEF`, a pattern `$$ABC_DEF` could result in value from either env vars.

- Added a fix to deterministically substitute the correct env variable.
- Added Unit tests for the same.

### DIFF
--- a/internal/config/substitutions.go
+++ b/internal/config/substitutions.go
@@ -113,23 +113,23 @@ func toString(value interface{}) string {
 // Use a double dollar sign eg. $$MY_ENV_VAR to subsitute that environment variable into the config file
 // Can be useful with kubernetes service environment variables
 func SubEnvVariables(strConf *string) {
-	subCount := strings.Count(*strConf, "$$")
-	replaceCount := 0
-	if subCount > 0 {
-		for _, e := range os.Environ() {
-			pair := strings.SplitN(e, "=", 2)
-			if len(pair) == 2 && pair[0] != "" {
-				if strings.Contains(*strConf, "$$"+pair[0]) {
-					*strConf = strings.Replace(*strConf, "$$"+pair[0], pair[1], -1)
-					load.StatusCounterIncrement("environmentVariablesReplaced")
-					replaceCount++
-				}
-			}
-			if replaceCount >= subCount {
-				break
-			}
+	envVars := os.Environ()
+	envMap := make(map[string]string, len(envVars))
+	for _, e := range envVars {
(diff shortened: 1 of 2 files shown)


ujwalsah (approved):
LGTM 🚀

internal/config/substitutions.go
```diff
@@ -113,23 +113,23 @@ func toString(value interface{}) string {
...
-			}
+	envVars := os.Environ()
+	envMap := make(map[string]string, len(envVars))
+	for _, e := range envVars {
+		pair := strings.SplitN(e, "=", 2)
+		if len(pair) == 2 && pair[0] != "" {
+			envMap[pair[0]] = pair[1]
 		}
 	}
+	re := regexp.MustCompile(`\$\$([A-Za-z_][A-Za-z0-9_]*)`)
```
sairaj18, line 124: By introducing this regex if there is an env variable `host-name` and input is `host-name is $$host-name` to this `SubEnvVariables()` func, will it substitue the env variable?

Do we need to re-think about this regex or add the syntax that should be used for env variables and document it somewhere?

sairaj18 (commented):
Left a comment

internal/config/substitutions.go
rajrohanyadav: > By introducing this regex if there is an env variable host-name and input is host-name is $$host-name to this SubEnvVariables() func, will it substitue the env variable?

Yes it will work fine.

> Do we need to re-think about this regex or add the syntax that should be used for env variables and document it somewhere?

Why do we need to re-think about the regex? I don't see a valid reason.
It is already well understood and documented. We had a bug in our implementation which this PR fixes.

rajrohanyadav: Please see the test case added in https://github.com/newrelic/nri-flex/pull/561/commits/b15f1624e639f303cdb8cbf8e3ed873585948812 for your clarity.

sairaj18: > Yes it will work fine.

I don't think so. Since the special character `-` is not present in the regexp

> Why do we need to re-think about the regex? I don't see a valid reason.

Special characters are not considered as mentioned above. Also can the env vars start with `_` or `-`?. So, I was asking if we need to document about the syntax they need to use why declaring env vars

> Please see the test case added in https://github.com/newrelic/nri-flex/commit/b15f1624e639f303cdb8cbf8e3ed873585948812 for your clarity.

In this test case u did not use `-`

PR #574, opened 2025-06-30
feat: add example for monitoring amd gpu using rocm-smi

rajrohanyadav:


examples/amd-rocm-smi-gpu-monitoring.yml
```diff
@@ -0,0 +1,107 @@
...
+  - name: nri-flex
+    # interval: 30s
+    config:
+      name: AMDRocmSMI
+      apis:
+        # Comprehensive AMD GPU metrics collection - all device details in single sample
+        - name: AMDGpuAllMetrics
+          event_type: "amd_gpu_metrics"
+          commands:
+            - run: "rocm-smi -ivftumcpS --showbus --showmemvendor --showdriverversion --showserial --showproductname --json | jq '(keys[0]) as $card | .[$card] + {card: $card}'"
```
rahulreddy15, line 25: Here `keys[0]` means we only get the first one right ?
What if there is more than one GPU.

examples/amd-rocm-smi-gpu-monitoring.yml
```diff
@@ -0,0 +1,107 @@
...
+            "GUID": "device_guid"
+            "VBIOS version": "vbios_version"
+            "Driver version": "driver_version"
+            "Serial Number": "serial_number"
+            "Card series": "card_series"
+            "Card model": "card_model"
+            "Card vendor": "card_vendor"
+            "Card SKU": "card_sku"
+            "Memory vendor": "memory_vendor"
+            "Temperature \\(Sensor edge\\) \\(C\\)": "temp_edge_c"
```
rahulreddy15, line 41: Would adding unit like `celsius` and `watt` make sense ?
Like `temp_memory_celsius` and `power_avg_graphics_watt`

examples/amd-rocm-smi-gpu-monitoring.yml
rahulreddy15, line 25: If `rocm-smi` or `jq` are not installed.
Will this command give the right error message.
Think it should ?

rahulreddy15 (changes requested):
Awesome job figuring this out. Left a few nits.

examples/amd-rocm-smi-gpu-monitoring.yml
rajrohanyadav: The keys already have units `_c` and `_w`. I don't think we need extra characters when a single char is enough to display the meaning. Saves bytes from ingestion cost.

rahulreddy15 (approved):
LGTM 🚀

PR #582, opened 2025-08-21
upgrading golang version to 1.24.6

rahulreddy15:


internal/inputs/netdial.go
```diff
@@ -53,7 +53,7 @@ func NetDialWithTimeout(dataStore *[]interface{}, command load.Command, dataSamp
 			dialError = err
 		} else {
 			if command.Run != "" {
-				fmt.Fprintf(dialConn, command.Run)
+				fmt.Fprint(dialConn, command.Run)
```
sairaj18, line 56: why is this change needed?

internal/config/secrets.go
```diff
@@ -43,7 +43,7 @@ func loadSecrets(config *load.Config) error {
 			err = fmt.Errorf("config: secret needs 'file', 'data' and 'http' parameter to be set")
 			load.Logrus.WithFields(logrus.Fields{
 				"secret": name,
-			}).Errorf(err.Error())
+			}).Error(err.Error())
```
sairaj18, line 46: Why is this change needed?

sairaj18 (commented):
Thanks for working on this change. Left few comments

internal/config/secrets.go
rahulreddy15: To fix this error in the tests:
Error: internal/config/secrets.go:47:14: non-constant format string in call to (*github.com/sirupsen/logrus.Entry).Errorf

internal/inputs/netdial.go
rahulreddy15: To fix this error in the tests:
Error: internal/inputs/netdial.go:57:27: non-constant format string in call to fmt.Fprintf

PR #617, opened 2026-02-25
fix: grant security-events write permission 

aryamankohli:
- Added 'security-events: write' to workflow permissions to enable uploading scan results to the GitHub Security tab.


muruganishekar:
addressed this change already

PR #624, opened 2026-03-19
examples(gcp): add BigQuery REST API integration example

dpacheconr:
## Summary

- Adds `bq-rest.yml` as a complementary BigQuery integration using the **BigQuery REST API** directly, alongside the existing `bq.yml` which uses the `bq` CLI
- Useful for environments where the `bq` CLI is not available (containers, restricted hosts)
- Uses the Flex `lookup` feature to chain an OAuth2 bearer token (fetched via `gcloud auth application-default print-access-token`) into subsequent HTTP API calls — gcloud caches the token locally so there is no network overhead on most polls
- Updates `README.md` with a comparison table and separate sections for each approach

## New events reported

| Event type | Description |
|---|---|
| `bq_rest_table_stats` | Lists all tables in a dataset with metadata via REST |
| `bq_rest_storage_stats` | Table storage stats via INFORMATION_SCHEMA query through the REST query API |

## Test plan

- [ ] Set `project_id` and `dataset_id` in `variable_store`
- [ ] Run `gcloud auth application-default login` on test host
- [ ] Validate manually: `[sudo] /opt/newrelic-infra/newrelic-integrations/bin/nri-flex --verbose --pretty --config_file bq-rest.yml`
- [ ] Confirm `bq_rest_table_stats` and `bq_rest_storage_stats` events appear in NRDB
- [ ] Confirm existing `bq.yml` is unchanged and still works

### DIFF
--- a/examples/GCP/BigQuery/README.md
+++ b/examples/GCP/BigQuery/README.md
@@ -1,23 +1,33 @@
-# Big Query Flex Integration
+# BigQuery Flex Integration
 
-This flex integration runs bq queries via GCP CLI and reports the output to New Relic. Most useful information is found under the [INFORMATION_SCHEMA](https://cloud.google.com/bigquery/docs/information-schema-intro) view.
+This directory contains two complementary Flex integrations for BigQuery, each using a different approach.
+
+| File | Approach | Best for |
+|---|---|---|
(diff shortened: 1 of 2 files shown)


muruganishekar (approved):
LGTM

PR #629, opened 2026-04-08
chore(deps): update docker from v28.5 to v29.3.1 and fix breaking cod…

gmanandhar-nr:
Updating docker from version 28.5 to v29.3.1

This is a major upgrade from docker where docker v29 split the monolithic module into separate sub-modules with independent version numbers. There is no single "v29.3.1" Go module. Instead, the fix is distributed across moby/client v0.4.0 and moby/api v1.54.1.

As part of this upgrade, there were multiple breaking changes such as change in return type, function signature, etc. This PR aims to adapt to the new API.

Below are the tests that were run to ensure the new code changes related to docker work (courtsey of claude :) )

## Prerequisites

**IMPORTANT:** nri-flex exits immediately if `flexConfigs/` doesn't
exist. You must create it before any test will work. If you cleaned
up from a previous run, redo ALL steps below.

```bash
# Build
go build -o bin/nri-flex ./cmd/nri-flex/

# Create required directories and dummy config
# Without this, nri-flex exits before reaching Docker code
mkdir -p flexConfigs flexContainerDiscovery
cat > flexConfigs/dummy.yml << 'EOF'
name: dummy
apis:
  - name: dummy
    commands:
      - run: echo '{"status":"ok"}'
EOF

# Start test containers (only docker commands in the entire guide)
docker run -d --name test-nginx nginx:alpine

docker run -d --name test-labeled \
  --label "flexDiscoveryRedis=t_redis.c_redis.tt_img.tm_contains" \
  redis:alpine

sleep 3

# Create discovery config that matches the labeled container
cat > flexContainerDiscovery/redis.yml << 'EOF'
name: redisDiscovery
apis:
  - name: redis
    entity: redis-${auto:ip}
    commands:
      - run: echo '{"ip":"${auto:ip}","port":"${auto:port}"}'
EOF
```

**Why these specific containers?**

- `test-nginx` — a plain container with no special labels. nri-flex
  will list it (ContainerList) and inspect it (ContainerInspect).
- `test-labeled` — has a `flexDiscoveryRedis` label. nri-flex will
  list it, inspect it, find the label, look up `redis.yml` in the
  discovery config directory, and attempt to build a dynamic config
  using the container's IP and port. This exercises the deepest
  code path.

---

## TEST 1: ContainerList()

**Code change:** `discovery.go:43-48`

```go
// Old (v28)
containerList, err := cli.ContainerList(ctx, container.ListOptions{})
len(containerList) > 0

// New (v29)
containerListResult, err := cli.ContainerList(ctx, client.ContainerListOptions{})
len(containerListResult.Items) > 0
```

**Run:**

```bash
./bin/nri-flex -container_discovery -container_dump -verbose 2>&1 \
  | grep -E "discovery: containers|container:"
```

**Expected output:**

```
discovery: containers 2 filtered containers 2
container: abc123 nginx:alpine [/test-nginx]
container: def456 redis:alpine [/test-labeled]
```

**How this proves the change works:**

- `"discovery: containers 2 filtered containers 2"` is logged at
  `discovery.go:136`, which is AFTER `ContainerList()` returns
  successfully at line 43 and `.Items` is iterated at lines 48-59.
- `"container: ..."` lines are logged at `discovery.go:140` inside a
  loop over the containers slice that came from `.Items`.
- If the v29 signature was wrong, you'd see `"unable to set perform
  container list"` instead, or a panic.

**Limitations:** None. This **fully tests** the ContainerList change
at runtime.

---

## TEST 2: ContainerInspect() with 3 parameters

**Code change:** `discovery.go:194` (and 314, 352, 403)

```go
// Old (v28)
containerInspect, err = dockerClient.ContainerInspect(ctx, container.ID)

// New (v29)
inspectResult, err := dockerClient.ContainerInspect(ctx, container.ID, client.ContainerInspectOptions{})
containerInspect = inspectResult.Container
```

**Run:**

```bash
./bin/nri-flex -container_discovery -container_dump -verbose 2>&1 \
  | grep -E "discovery: (fwd lookup|container inspect failed|containers)"
```

**Expected output:**

```
discovery: containers 2 filtered containers 2
discovery: fwd lookup for flexDiscoveryRedis
```

**How this proves the change works:**

- `runForwardLookup()` calls `ContainerInspect()` at line 194 for
  **every** discovered container. If the 3-parameter call failed, you'd
  see `"discovery: container inspect failed"` at line 201.
- `"discovery: fwd lookup for flexDiscoveryRedis"` is logged at line
  225. This line is only reached AFTER:
  1. `ContainerInspect()` succeeds (line 194)
  2. `.Container` is extracted from the result (line 196)
  3. `.Config` is not nil (line 207)
  4. Labels are read from `.Config.Labels` (line 209)
  5. The `flexDiscoveryRedis` key is found in the labels (line 223)
- So seeing this log line proves the entire inspect-and-read-labels
  chain works.

**Limitations:** None. This **fully tests** the ContainerInspect
change at runtime. The forward lookup runs ContainerInspect on every
container automatically.

---

## TEST 3: Network IP extraction via netip.Addr.String()

**Code change:** `discovery.go:458`

```go
// Old (v28)
networkIPAddress = network.IPAddress   // was a string

// New (v29)
networkIPAddress = network.IPAddress.String()   // now netip.Addr
```

**Run:**

```bash
./bin/nri-flex -container_discovery -verbose 2>&1 \
  | grep -E "discovery: .*(redis\.yml|ip|port|matched|couldn)"
```

**Expected output (one of these):**

If IP extraction and config matching succeed:
```
discovery: abc123 matched redis.yml
discovery: abc123 redis.yml - private - 172.17.0.3:6379
```

If IP is found but config doesn't fully match:
```
discovery: abc123 couldn't build dynamic cfg
discovery: abc123 missing variable unable to create dynamic cfg ip:172.17.0.3 - port:
```

**How this proves the change works:**

- Line 458 (`network.IPAddress.String()`) runs inside
  `addDynamicConfig()`, which is called when the forward lookup
  matches a container label to a discovery config file.
- The log at line 522 (`"discovery: <id> <file> - <mode> - <ip>:<port>"`)
  shows the extracted IP, which could only have come from calling
  `.String()` on the `netip.Addr` field.
- Even if the full config doesn't build, the log at line 530 still
  prints the IP value, proving extraction worked.
- If `.String()` was wrong (type mismatch), the binary wouldn't
  compile at all.

**Limitations:** This code path only executes when a container's
`flexDiscovery` label matches a file in `flexContainerDiscovery/`.
That's why the prerequisites create `test-labeled` with the
`flexDiscoveryRedis` label AND the matching `redis.yml` file. Without
both, this line doesn't run at runtime. The test setup above is
designed to trigger this path.

---

## TEST 4: string(targetContainer.State)

**Code change:** `discovery.go:563`

```go
// Old (v28)
yml.CustomAttributes["container.state"] = targetContainer.State   // was string

// New (v29)
yml.CustomAttributes["container.state"] = string(targetContainer.State)   // now ContainerState type
```

**How to verify:**

This line runs inside `addDynamicConfig()` in the same code block as
TEST 3. If TEST 3 shows a log like `"discovery: <id> redis.yml -
private - <ip>:<port>"` (line 522) WITHOUT subsequent errors, it means
the code continued past line 563 successfully.

**Limitations:** Same as TEST 3 — requires a matching discovery
config. Additionally, there is no dedicated log message for this
specific line. If it failed with a type error, it would cause a
compilation failure (which we know doesn't happen since the binary
builds). If it panicked at runtime, you'd see it in the logs.

---

## TEST 5: ExecCreate() and ExecAttach()

**Code change:** `docker.go:54, 63`

```go
// Old (v28)
exec, err := cli.ContainerExecCreate(ctx, containerID, execConfig)
containerConn, err := cli.ContainerExecAttach(ctx, exec.ID, execAttachConfig)

// New (v29)
execResult, err := cli.ExecCreate(ctx, containerID, execConfig)
attachResult, err := cli.ExecAttach(ctx, execResult.ID, execAttachConfig)
```

**How to verify:**

```bash
./bin/nri-flex -container_discovery -verbose 2>&1 \
  | grep -E "hostname.*fallback|exec|low level ip"
```

**Limitations:** This is the **hardest change to test at runtime**.
`ExecCreate()` and `ExecAttach()` are only called from
`execHostnameFallback()` (discovery.go:699), which is a last-resort
fallback when a container's IP cannot be found through:

1. Docker network settings (the normal path)
2. Low-level `/proc/<pid>/net/fib_trie` reading

In a normal Docker setup, the network IP is always available, so
this fallback never triggers. **This code path cannot be reliably
triggered in a standard local test.**

**What we CAN verify:**

1. The binary compiles — proving the renamed methods and new option
   types are type-correct:
   ```bash
   grep -n "cli.ExecCreate\|cli.ExecAttach" internal/discovery/docker.go
   ```
   Expected:
   ```
   54: execResult, err := cli.ExecCreate(ctx, containerID, execConfig)
   63: attachResult, err := cli.ExecAttach(ctx, execResult.ID, execAttachConfig)
   ```

2. The old method names are gone:
   ```bash
   grep "ContainerExecCreate\|ContainerExecAttach" internal/discovery/docker.go
   ```
   Expected: no output (old names removed).

3. The unit test passes:
   ```bash
   go test ./internal/discovery/... -v
   ```

**Honest assessment:** We cannot trigger `ExecCreate`/`ExecAttach` at
runtime in a normal local test. Compilation proves type correctness.
A full runtime test would require a container whose IP is not
discoverable through network settings — an unusual scenario typically
seen in complex Kubernetes or custom network setups.

---

## Summary: What Each Test Actually Proves

| Test | Code Change | Fully Tested at Runtime? | How |
|------|-------------|--------------------------|-----|
| **TEST 1** | `ContainerList()` returns `ContainerListResult.Items` | **Yes** | Log shows container count and dump |
| **TEST 2** | `ContainerInspect()` takes 3 params, returns `.Container` | **Yes** | Log shows labels read from inspect result |
| **TEST 3** | `network.IPAddress.String()` | **Yes** (with setup) | Log shows extracted IP; requires matching discovery config |
| **TEST 4** | `string(targetContainer.State)` | **Partially** | Same code path as TEST 3; no dedicated log line; compilation proves type correctness |
| **TEST 5** | `ExecCreate()` / `ExecAttach()` | **No** (compilation only) | Fallback path that requires unreachable-IP containers; compilation proves type correctness |

---

## Cleanup

```bash
docker stop test-nginx test-labeled 2>/dev/null
docker rm test-nginx test-labeled 2>/dev/null
rm -rf flexContainerDiscovery
```

---


muruganishekar (approved):
LGTM, please scan using trivy before merging it.

gmanandhar-nr:
trivy not showing any vulnerabilities anymore
image

PR #642, opened 2026-05-27
fix: resolve CodeQL go/insecure-hostkeycallback in SCP

ujwalsah:
## Summary
- Replace `ssh.InsecureIgnoreHostKey()` with proper known_hosts verification using `knownhosts.New`
- Added `known_hosts_file` config option to specify a custom known_hosts path (supports `~` expansion)
- Returns actionable error if known_hosts file is not found
- Resolves CodeQL alert `go/insecure-hostkeycallback` ([NR-560141](https://new-relic.atlassian.net/browse/NR-560141))

## Breaking change — no opt-out

This PR removes all insecure SSH host key handling. There is **no fallback** and **no opt-out option**:

- Every SCP target host **must** be present in a known_hosts file (default: `~/.ssh/known_hosts`)
- Users on dynamic/ephemeral infrastructure must provision known_hosts as part of their setup, e.g.:
  ```
  ssh-keyscan <target-host> >> ~/.ssh/known_hosts
  ```
- Alternatively, specify a custom file with the `known_hosts_file` config option

Users who previously relied on the insecure behavior will see connection errors until they add their target hosts to known_hosts.

## Test plan
- [x] Unit tests for `getHostKeyCallback` (missing file, custom path, default path)
- [x] Unit tests for `expandHome` (tilde expansion)
- [x] `go test ./internal/inputs/...` passes
- [ ] Smoke test against a real host in known_hosts to verify end-to-end connection
- [ ] Confirm CodeQL alert auto-closes after merge

[NR-560141]: https://new-relic.atlassian.net/browse/NR-560141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/inputs/scp.go
+++ b/internal/inputs/scp.go
@@ -10,6 +10,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -17,6 +19,7 @@ import (
 	"github.com/newrelic/nri-flex/internal/utils"
 	"github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/knownhosts"
 )
 
 // RunScpWithTimeout performs scp with timeout to gather data from a remote file.
@@ -68,14 +71,19 @@ func getSSHConnection(yml *load.Config, api load.API) (*sftp.Client, error) {
 		return nil, err
 	}
 
+	hostKeyCallback, err := getHostKeyCallback(api.Scp)
+	if err != nil {
+		return nil, fmt.Errorf("ssh: failed to set up host key verification: %v", err)
+	}
+
 	sshConfig := &ssh.ClientConfig{
 		User:            user,
-		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		HostKeyCallback: hostKeyCallback,
 		Timeout:         timeout,
 		Auth: []ssh.AuthMethod{
 			authMethod,
 		},
-	} // #nosec
+	}
 
 	sshConfig.SetDefaults()
 
@@ -90,6 +98,44 @@ func getSSHConnection(yml *load.Config, api load.API) (*sftp.Client, error) {
 		return nil, fmt.Errorf("ssh: failed to init sftp client, error: %v", err)
 	}
 	return client, nil
+}
+
+func getHostKeyCallback(scp load.SCP) (ssh.HostKeyCallback, error) {
+	knownHostsFile := expandHome(scp.KnownHostsFile)
+	if knownHostsFile == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return nil, fmt.Errorf("unable to determine home directory: %v", err)
+		}
+		knownHostsFile = filepath.Join(home, ".ssh", "known_hosts")
+	}
+
+	if _, err := os.Stat(knownHostsFile); os.IsNotExist(err) {
+		return nil, fmt.Errorf("known_hosts file not found at %s. Create the file with "+
+			"'ssh-keyscan <host> >> %s', or specify a custom path with 'known_hosts_file'",
+			knownHostsFile, knownHostsFile)
+	}
+
+	callback, err := knownhosts.New(knownHostsFile)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read known_hosts file %s: %v", knownHostsFile, err)
+	}
+	return callback, nil
+}
+
+// expandHome replaces a leading ~ with the user's home directory.
+func expandHome(path string) string {
+	if path == "" {
+		return ""
+	}
+	if strings.HasPrefix(path, "~/") {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return path
+		}
+		return filepath.Join(home, path[2:])
+	}
+	return path
 }
 
 func getAuthMethod(yml *load.Config, api load.API) (ssh.AuthMethod, error) {
--- a/internal/inputs/scp_test.go
+++ b/internal/inputs/scp_test.go
@@ -6,11 +6,97 @@
 package inputs
 
 import (
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/newrelic/nri-flex/internal/load"
 )
+
+func TestGetHostKeyCallback_MissingKnownHostsFile(t *testing.T) {
+	scp := load.SCP{KnownHostsFile: "/nonexistent/path/known_hosts"}
+	_, err := getHostKeyCallback(scp)
+	if err == nil {
+		t.Fatal("expected error for missing known_hosts file")
+	}
+	if !strings.Contains(err.Error(), "known_hosts file not found") {
+		t.Fatalf("expected 'known_hosts file not found' error, got: %v", err)
+	}
+}
+
+func TestGetHostKeyCallback_CustomKnownHostsFile(t *testing.T) {
+	tmpDir, err := os.MkdirTemp("", "scp-test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	// Create a valid (empty) known_hosts file
+	knownHostsPath := filepath.Join(tmpDir, "known_hosts")
+	if err := os.WriteFile(knownHostsPath, []byte(""), 0600); err != nil {
+		t.Fatal(err)
+	}
+
+	scp := load.SCP{KnownHostsFile: knownHostsPath}
+	callback, err := getHostKeyCallback(scp)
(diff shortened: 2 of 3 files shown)


internal/inputs/scp.go
```diff
@@ -68,14 +71,19 @@ func getSSHConnection(yml *load.Config, api load.API) (*sftp.Client, error) {
...
 
+	hostKeyCallback, err := getHostKeyCallback(api.Scp)
+	if err != nil {
+		return nil, fmt.Errorf("ssh: failed to set up host key verification: %v", err)
+	}
+
 	sshConfig := &ssh.ClientConfig{
 		User:            user,
-		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		HostKeyCallback: hostKeyCallback,
```
shashank-reddy-nr, line 81: Consider setting `HostKeyAlgorithms` to avoid handshake failures with knownhosts.
The knownhosts callback rejects the handshake when the server offers a host-key type that differs from the one stored in `known_hosts (ssh: handshake failed: knownhosts: key mismatch / no common algorithm for host key). SetDefaults()` doesn't fix this. The robust pattern:


`sshConfig.HostKeyAlgorithms = knownhosts.HostKeyAlgorithms(hostKeyCallback, addr)`

At minimum, please test against a real host whose key is already in known_hosts to confirm the happy path works.

internal/inputs/scp.go
```diff
@@ -92,6 +100,36 @@ func getSSHConnection(yml *load.Config, api load.API) (*sftp.Client, error) {
...
 
+func getHostKeyCallback(scp load.SCP) (ssh.HostKeyCallback, error) {
+	if scp.SkipHostKeyVerification {
+		load.Logrus.Warn("SCP: host key verification is disabled (skip_host_key_verification: true). " +
+			"This is insecure and should only be used in trusted networks. " +
+			"Consider configuring known_hosts_file instead.")
+		return ssh.InsecureIgnoreHostKey(), nil //nolint:gosec // User explicitly opted in
+	}
+
+	knownHostsFile := scp.KnownHostsFile
```
shashank-reddy-nr: `~` in a user-supplied path won't be expanded. If someone sets known_hosts_file: `~/.ssh/known_hosts`, Go passes the literal ~ to `os.Stat/knownhosts.New` and it fails with a confusing "not found". Either expand a leading ~ here, or document that known_hosts_file must be an absolute path. (The default branch is fine since it uses os.UserHomeDir().)

internal/inputs/scp.go
```diff
@@ -92,6 +100,36 @@ func getSSHConnection(yml *load.Config, api load.API) (*sftp.Client, error) {
 	return client, nil
 }
 
+func getHostKeyCallback(scp load.SCP) (ssh.HostKeyCallback, error) {
+	if scp.SkipHostKeyVerification {
+		load.Logrus.Warn("SCP: host key verification is disabled (skip_host_key_verification: true). " +
```
shashank-reddy-nr: this warning fires on every `SCP collection. getSSHConnection` runs each interval, so with skip_host_key_verification: true the warning logs repeatedly. Consider logging it once (e.g. sync.Once) or downgrading to debug after the first emit. Acceptable as-is if SCP runs are infrequent.

internal/load/load.go
```diff
@@ -498,13 +498,15 @@ type JMX struct {
...
-	SSHPEMFile string `yaml:"ssh_pem_file"`
+	User                     string `yaml:"user"`
+	Pass                     string `yaml:"pass"`
+	Host                     string `yaml:"host"`
+	Port                     string `yaml:"port"`
+	RemoteFile               string `yaml:"remote_file"`
+	Passphrase               string `yaml:"pass_phrase"`
+	SSHPEMFile               string `yaml:"ssh_pem_file"`
+	KnownHostsFile           string `yaml:"known_hosts_file"`
+	SkipHostKeyVerification  bool   `yaml:"skip_host_key_verification"`
```
shashank-reddy-nr: Run `gofmt/goimports` on this block. Adding the two longer field names shifts the alignment column for the whole SCP struct - just confirm `gofmt -l ./internal/load` reports nothing so CI doesn't fail on formatting.

internal/inputs/scp_test.go
```diff
@@ -97,6 +97,8 @@ func TestGetSSHConnection(t *testing.T) {
 			"no route to host",
 			"network is unreachable",
 			"failed to connect to sftp host",
+			"known_hosts file not found",
```
shashank-reddy-nr, line 186: The new getHostKeyCallback logic isn't directly tested. This only widens the accepted-error list of the existing live-host test. getHostKeyCallback is pure and easy to cover — please add unit tests for:

- `skip_host_key_verification: true `→ non-nil callback, no error
- missing known_hosts_file → error containing `"known_hosts file not found"`
- custom path pointed at a temp file with a valid `known_hosts` line → working callback
- empty path → falls back to `~/.ssh/known_hosts`

internal/inputs/scp.go
```diff
@@ -92,6 +100,36 @@ func getSSHConnection(yml *load.Config, api load.API) (*sftp.Client, error) {
...
+			"This is insecure and should only be used in trusted networks. " +
+			"Consider configuring known_hosts_file instead.")
+		return ssh.InsecureIgnoreHostKey(), nil //nolint:gosec // User explicitly opted in
```
shashank-reddy-nr: This probably won't keep the CodeQL alert closed — `//nolint:gosec` doesn't apply to CodeQL.
nolint directives are only read by `golangci-lint/gosec;` CodeQL ignores them. Since `ssh.InsecureIgnoreHostKey()` still exists here and CodeQL traces the returned value into the HostKeyCallback assignment, `go/insecure-hostkeycallback` will very likely re-fire on this line even after the original line is gone.

internal/inputs/scp.go
shashank-reddy-nr: Can you please check this as well?

internal/inputs/scp.go
```diff
@@ -92,6 +106,57 @@ func getSSHConnection(yml *load.Config, api load.API) (*sftp.Client, error) {
...
+				"This is insecure and should only be used in trusted networks. " +
+				"Consider configuring known_hosts_file instead.")
+		})
+		// Custom callback that accepts all host keys. This avoids calling
+		// ssh.InsecureIgnoreHostKey() directly which triggers CodeQL alerts.
+		return func(_ string, _ net.Addr, _ ssh.PublicKey) error {
+			return nil
+		}, nil
+	}
+
```
shashank-reddy-nr: This silences CodeQL but doesn't change the security posture — please confirm this is the agreed resolution.
This callback is byte-for-byte what `ssh.InsecureIgnoreHostKey()` does internally (unconditional return nil), so the skip_host_key_verification path is exactly as insecure as before — the only change is that the scanner can't see it. The comment effectively documents scanner evasion.
Two asks: (1) verify on an actual CodeQL run that the alert clears (some versions of g`o/insecure-hostkeycallback` also flag trivial always-nil callbacks); (2) check with whoever owns the alert whether this is acceptable vs. keeping readable code (e.g. `ssh.InsecureIgnoreHostKey()`) and formally dismissing the alert in the Security tab with justification. Functionally fine - this is an optics/process call.

internal/inputs/scp.go
shashank-reddy-nr, line 81: known_hosts is now the only path - please guard against the host-key-algorithm handshake gotcha.
`knownhosts.New` fails the handshake when the server's offered host-key type differs from the one stored in `known_hosts` (knownhosts: key mismatch / no common algorithm for host key), and `SetDefaults()` doesn't fix it. Since there's no insecure fallback anymore, set the algorithms from the stored key:


```
addr := net.JoinHostPort(host, port)
sshConfig.HostKeyAlgorithms = knownhosts.HostKeyAlgorithms(hostKeyCallback, addr)
```
(Requires re-adding the net import and reusing addr in `ssh.Dia`l) At minimum, please smoke-test against a real host already in known_hosts to confirm the verified path connects end-to-end - this is the one runtime risk left.

shashank-reddy-nr:
@ujwalsah The new approach is the right call - removing `InsecureIgnoreHostKey()` and the skip option entirely means CodeQL genuinely clears instead of being suppressed.

Two things before merge:

- The breaking change is now stricter than the description may suggest - there is no longer any opt-out, so every SCP user must have the target host in a known_hosts file or collection fails. Please make sure the PR description and CHANGELOG reflect the final no-fallback behavior, and flag that users on dynamic/ephemeral infra will need to ssh-keyscan hosts into provisioning.
- Verify the verified path actually connects against a real host in known_hosts (see the inline note on HostKeyAlgorithms) - that's the only untested runtime risk.

shashank-reddy-nr (approved):

PR #646, opened 2026-05-29
fix: use PBKDF2 for key derivation to resolve CodeQL alert

ujwalsah:
## Summary
- Replace SHA256 direct hashing with PBKDF2 (100k iterations, SHA256) for AES key derivation
- CodeQL's `go/weak-sensitive-data-hashing` rule requires a computationally expensive function for sensitive data
- SHA256 alone is considered insufficient as it's not designed to be slow
- Resolves CodeQL alert #17 ([NR-560139](https://new-relic.atlassian.net/browse/NR-560139))

## Breaking change
Data encrypted with the previous version (MD5 or SHA256 key derivation) **cannot be decrypted** with this version since the key derivation has changed to PBKDF2.

## Test plan
- [ ] Verify CI passes
- [ ] Confirm CodeQL alert #17 auto-closes after merge

[NR-560139]: https://new-relic.atlassian.net/browse/NR-560139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/utils/simpleEncryt.go
```diff
@@ -10,22 +10,22 @@ import (
 	"crypto/cipher"
 	"crypto/rand"
 	"crypto/sha256"
-	"encoding/hex"
 	"io"
+
+	"golang.org/x/crypto/pbkdf2"
```
shashank-reddy-nr: `golang.org/x/crypto/pbkdf2` is deprecated in favor of the `stdlib crypto/pbkdf2`, added in Go 1.24. This module is on go 1.26.3, so you could use `crypto/pbkdf2`

internal/utils/simpleEncryt.go
```diff
@@ -10,22 +10,22 @@ import (
...
 )
 
-func createHash(key string) string {
-	hasher := sha256.New()
-	hasher.Write([]byte(key))
-	return hex.EncodeToString(hasher.Sum(nil))
+func deriveKey(passphrase string) []byte {
```
shashank-reddy-nr: There is no simpleEncryt_test.go in the folder. Adding a test case would be better so that, we can test before breaking on master branch. Consider a positive and negative test case.

shashank-reddy-nr (dismissed):

shashank-reddy-nr:
@ujwalsah Changes looks good. Before merge: 
- add the `len(data) < nonceSize` guard in Decrypt
- add a comment justifying the static salt (or switch to a prepended random salt)
- rebase to clear the conflict. The breaking-change note is good, please also make sure it lands in the CHANGELOG/release notes, since existing encrypted secrets will need re-encryption after upgrade.

shashank-reddy-nr:
Please resolve the merge conflicts by rebasing the PR.

shashank-reddy-nr (approved):
LGTM

